PR #43, opened 2026-04-24
fix(model): handle numeric component prop values without crashing

### LINKED ISSUE #42: lint: numeric fontWeight on components crashes model builder with "raw.match is not a function"
## Summary

The `lint` command crashes with `Unexpected error during model building: raw.match is not a function` when a component defines `fontWeight` as a bare number (e.g. `fontWeight: 600`). Quoting the value (`fontWeight: "600"`) avoids the crash.

## Spec conflict

`docs/spec.md` explicitly states both forms should be equivalent:

> `fontWeight` (number) - A numeric font weight value (e.g., `400`, `700`). In YAML, this may be expressed as either a bare number or a quoted string; both are equivalent.

This contract is broken in the `components` block (typography block handles numeric weights fine).

## Minimal repro

```yaml
---
version: "alpha"
name: Repro
colors:
  primary: "#000000"
typography:
  h1:
    fontFamily: "Inter"
    fontSize: 3rem
rounded:
  md: 8px
spacing:
  md: 16px
components:
  button-primary:
    backgroundColor: "{colors.primary}"
    fontWeight: 600      # <-- crashes here
---
```

Lint output:

```json
{
  "findings": [{
    "severity": "error",
    "message": "Unexpected error during model building: raw.match is not a function"
  }],
  "summary": { "errors": 1, "warnings": 0, "infos": 0 }
}
```

Quoting the value (`fontWeight: "600"`) makes the file lint cleanly. Same crash occurs for any other bare-number value on a component prop (e.g. `borderWidth: 1`).

## Environment

- `@google/design.md@0.1.1` via `npx`
- macOS 15.4 (Darwin 25.4.0), Node 20
- Reproducible in an empty directory

## Suggested fix

Coerce the value to string (or check `typeof raw === 'string'` before `.match`) in the component-token model builder — same handling the typography parser already uses.

Happy to PR a test case if useful.

tejas100:
Fixes #42

## Problem
The `lint` command crashes with `Unexpected error during model building: 
raw.match is not a function` when a component defines a numeric prop like 
`fontWeight: 600` or `borderWidth: 1`. The spec explicitly states bare 
numbers and quoted strings are equivalent, so this is a spec violation.

## Root Cause
In Phase 3 of `ModelHandler`, all component prop values were passed 
directly to `isTokenReference()` and `isValidColor()`, both of which call 
`.match()` internally. When `rawValue` is a number, `.match` doesn't exist 
— crash.

## Fix
Added a `typeof rawValue === 'number'` guard at the top of the component 
property loop, storing numeric values as-is. This mirrors the pattern 
already used by `parseTypography` for the same property.

## Tests
Added 4 new test cases covering:
- `fontWeight: 600` (exact repro from issue)
- `fontWeight: 700` stored correctly as a number
- `borderWidth: 1` (also mentioned in issue)
- Mixed numeric + string + token ref props in one component

### DIFF
--- a/packages/cli/src/linter/model/handler.test.ts
+++ b/packages/cli/src/linter/model/handler.test.ts
@@ -356,4 +356,66 @@ describe('ModelHandler', () => {
       }
     });
   });
+
+  // ── Fix #42: numeric component props crash model builder ──────────
+  describe('numeric component property values', () => {
+    it('does not crash when fontWeight is a bare number', () => {
+      const result = handler.execute(makeParsed({
+        colors: { primary: '#000000' },
+        components: {
+          'button-primary': {
+            backgroundColor: '{colors.primary}',
+            fontWeight: 600 as unknown as string,
+          },
+        },
+      }));
+      expect(result.findings.filter(f => f.severity === 'error')).toHaveLength(0);
+      const btn = result.designSystem.components.get('button-primary');
+      expect(btn).toBeDefined();
+      expect(btn?.properties.get('fontWeight')).toBe(600);
+    });
+
+    it('stores numeric fontWeight value as-is in component properties', () => {
+      const result = handler.execute(makeParsed({
+        components: {
+          'heading': {
+            fontWeight: 700 as unknown as string,
+          },
+        },
+      }));
+      const heading = result.designSystem.components.get('heading');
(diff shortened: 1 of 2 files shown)


davideast:
Thank you @tejas100! I'm working on getting around to a lot of these and I'll review this soon.

davideast (approved):

PR #44, opened 2026-04-24
feat(spec): add Iconography section and icons.* tokens

### LINKED ISSUE #41: Feature: add Iconography section and `icons.*` tokens
## Summary

Add a formal `## Iconography` section and an `icons.*` token group to the
DESIGN.md spec. The current spec already uses `## Iconography` as its
illustrative example of a valid-but-undefined unknown section (`docs/spec.md`,
"Consumer Behavior for Unknown Content" table); this proposal fills that gap.

## Motivation

Every example DESIGN.md in the repo — `paws-and-paths`, `atmospheric-glass`,
`totality-festival` — describes icons in prose but has no token slot for them.
Agents consuming these files today have to infer the icon library, style
variant, and size scale from surrounding context. That defeats the "consistent
visual identity across tools" goal: two agents reading the same DESIGN.md
will pick different icon sets.

Iconography is a visual-language primitive alongside Colors, Typography,
Elevation, and Shapes, and belongs in the token schema.

### Token group

```yaml
icons:
  library: "Lucide"              # string — required within this group
  style: "outlined"              # enum (optional): outlined | filled | rounded | sharp | duotone
  strokeWidth: 1.5               # number (optional) — stroke-based libraries (Lucide, Tabler, Feather); font-based libraries map agent-side
  grid: 24px                     # Dimension (optional) — source viewBox unit; e.g. Radix=15, Phosphor=256, most others=24
  size:                          # map<string, Dimension> (optional)
    sm: 16px
    md: 24px
    lg: 32px
  color: "{colors.on-surface}"   # Color | token ref (optional)
```

### Section placement

New `## Iconography` section inserted between `Shapes` and `Components` in
the canonical section order. Rationale: it sits with the other visual-language
primitives (Colors, Typography, Elevation, Shapes) and immediately precedes
Components, where icon tokens are consumed.

### Prose example

```markdown
## Iconography

Icons use **Lucide** in its outlined style at a 1.5 stroke weight, echoing
the calm, linear rhythm of the typography. Icons render on a 24px grid and
inherit `on-surface` color by default.

- **Small (16px):** Inline with body text or dense tables.
- **Medium (24px):** Default for buttons, navigation, and form fields.
- **Large (32px):** Feature tiles and empty states.
```

### Recommended token names (non-normative)

Add two entries to the existing "Recommended Token Names" list, matching its
existing paragraph format:

**Icon Styles:** `outlined`, `filled`, `rounded`, `sharp`, `duotone`

**Icon Sizes:** `sm`, `md`, `lg`

## Design decisions

- **`library` is a free-form string** — the icon ecosystem is long-tail and
  new libraries appear constantly; an enum would calcify.
- **`style` enum covers five canonical values** drawn from a survey of
  Material Symbols, Heroicons, Phosphor, Font Awesome, Tabler, Lucide,
  Ionicons, Carbon, and Fluent. Unknown values accepted with warning,
  matching the spec's existing behavior.
- **`strokeWidth` is the cross-library weight concept.** Font-based libraries
  (e.g., Material Symbols weight 100–700) map it agent-side.
- **`color` accepts token refs** so icons participate in theming automatically
  as soon as #13 (themes) lands.
- **`size` mirrors `spacing`'s map-of-named-steps pattern** for consistency
  with existing token group conventions.
- **`grid` describes the source viewBox**, not the render size. This
  disambiguates "what unit is `strokeWidth` in?" for libraries on non-24
  grids (Radix 15, Phosphor 256).

## Deferred to follow-up

Each of these is a real case worth solving, but each is also its own design
debate. Deferring keeps v1 focused on the 80% case:

- **Duotone `secondaryColor`** — only Phosphor Duotone and FA Duotone need
  it; adding a union-typed `color` field or a sibling `secondaryColor` is
  worth discussing in isolation.
- **Separate `weight` axis vs unified `strokeWidth`** — Material Symbols'
  100–700 weight range may not map cleanly to a single stroke number; worth
  empirical validation.
- **Multi-library map** — real pattern (Lucide UI + Simple Icons for brand
  logos); prose handles it in v1.
- **Optical-size variant selection** — Carbon, Fluent, and Material Symbols
  ship different icon files per render size; v1 treats `size.*` as CSS
  output only, agent picks the right source file if the library supports
  optical sizing.

## Library-specific constraints (not schema concerns)

Some libraries couple fields in ways the schema can't express:

- **Heroicons** — "mini" is always solid 20px; "micro" is always solid 16px.
  `style: "outlined"` + `size.sm: 16px` is invalid for Heroicons specifically.
- **Material Symbols** — the fill/weight/grade/optical-size axes are
  library-specific and resolved agent-side.

These constraints live at the agent/library boundary, not in the linter —
the schema stays library-agnostic.

## Backwards compatibility

Purely additive. Existing DESIGN.md files without an `icons` block or
`## Iconography` section parse and lint identically. No version bump
required (spec is in `alpha`; additive schema changes are in-scope).

## Consumer-side impact

- `## Iconography` is used in `docs/spec.md` as the example unknown section.
  When this lands, that example row should be swapped (e.g., to
  `## Illustrations`) to avoid self-contradiction.
- `export` commands (Tailwind, DTCG): neither format has an icon token
  concept; v1 passes icon tokens through unchanged rather than emitting them.
- `import` command (#40): Tailwind configs and DTCG files do not carry icon
  metadata; import is a no-op for icons.

dovas-net:
# feat(spec): add Iconography section and `icons.*` tokens

This PR adds an `## Iconography` section and `icons.*` token group to the
DESIGN.md spec, so design systems can formally declare icon library, style
variant, stroke weight, grid, size scale, and default color.

Fixes #41

## What's added

- **Spec section** — new `## Iconography` inserted between `Shapes` and
  `Components` in the canonical section order.
- **Token group** — `icons.library`, `icons.style`, `icons.strokeWidth`,
  `icons.grid`, `icons.size.*`, `icons.color`.
- **Recommended token names** — `sm | md | lg` size scale,
  `outlined | filled | rounded | sharp | duotone` style values, added
  non-normatively to the existing recommended names list.
- **Linter** — enum validation on `style` (warn on unknown, matching existing
  spec behavior), Dimension validation on `size.*` and `grid`, number
  validation on `strokeWidth`, cross-reference validation on `color` reusing
  the existing token-ref resolver in `packages/cli/src/linter/linter/`.

## Files

- `packages/cli/src/linter/spec-config.yaml` — source-of-truth schema addition
- `packages/cli/src/linter/spec-config.test.ts` — schema tests
- `packages/cli/src/linter/model/` — type definitions for the `icons` group
- `packages/cli/src/linter/parser/` — YAML frontmatter parsing for `icons`
- `packages/cli/src/linter/fixtures/` — valid and invalid fixtures
- `docs/spec.md` — regenerated via `bun run spec:gen`
- `examples/paws-and-paths/DESIGN.md` — demo usage on one existing example
- **No changes** to `packages/cli/src/linter/spec-gen/`; the icons section
  renders via the existing section template.

## Consumer Behavior table update

`docs/spec.md`'s "Consumer Behavior for Unknown Content" table previously
used `## Iconography` as its example of an unknown section. Since that
section is now defined, the example row has been swapped to
`## Illustrations` to avoid self-contradiction.

## Backwards compatibility

No breaking changes. Existing DESIGN.md files without an `icons` block or
`## Iconography` section parse and lint identically. No spec version bump
— the spec is in `alpha` and additive schema changes are in-scope.

## Out of scope for this PR

Duotone `secondaryColor`, multi-library maps, separate `weight` axis, and
optical-size variant selection are deferred; see the parent issue for the
per-case rationale.

Icon tokens are spec-only in v1: `export` (Tailwind, DTCG) passes icon
tokens through unchanged, since neither format carries icon metadata.
`import` (#40) is a no-op for icons for the same reason.

## Testing

- Unit tests cover: valid icon tokens, invalid `style` enum value (warning
  only), unresolved `color` token reference (error), malformed `size.*`
  Dimension, malformed `grid` Dimension, non-numeric `strokeWidth`, and all
  optional fields omitted.
- Integration: `paws-and-paths` with a populated `icons` block lints clean.
- Export commands verified to pass `icons.*` tokens through unchanged on the
  updated example (no Tailwind/DTCG surface for icons in v1).

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -87,8 +87,9 @@ Every `DESIGN.md` follows the same structure. Sections can be omitted if they're
 4. **Layout** (also: "Layout & Spacing")
 5. **Elevation & Depth** (also: "Elevation")
 6. **Shapes**
-7. **Components**
-8. **Do's and Don'ts**
+7. **Iconography**
+8. **Components**
+9. **Do's and Don'ts**
 
 ### Prose and Tokens
 
@@ -274,6 +275,50 @@ rounded:
   full: 9999px
 ```
 
+## Iconography
+
+This section defines the icon library, style, and size scale used across the product.
+
+Example:
+
+```markdown
+## Iconography
+
+Icons use **Lucide** in its outlined style at a 1.5 stroke weight, echoing
+the calm, linear rhythm of the typography. Icons render on a 24px grid and
+inherit `on-surface` color by default.
+
+- **Small (16px):** Inline with body text or dense tables.
+- **Medium (24px):** Default for buttons, navigation, and form fields.
+- **Large (32px):** Feature tiles and empty states.
+```
+
+### Design Tokens
+
+The `icons` section defines a single token group (not a map) describing the
+icon library in use, its style variant, stroke weight, source grid, render
+sizes, and default color.
+
+```yaml
+icons:
+  library: "Lucide"
+  style: "outlined"
+  strokeWidth: 1.5
+  grid: "24px"
+  size:
+    sm: "16px"
+    md: "24px"
+    lg: "32px"
+  color: "{colors.on-surface}"
+```
+
+* `library` (string) — the icon set in use (e.g., "Lucide", "Heroicons", "Material Symbols"). Agents resolve the string to an import.
+* `style` (enum, optional) — one of: `outlined`, `filled`, `rounded`, `sharp`, `duotone`. Unknown values are preserved with a warning.
+* `strokeWidth` (number, optional) — for stroke-based libraries (Lucide, Tabler, Feather). Font-based libraries (e.g., Material Symbols with its 100–700 weight axis) map this agent-side.
+* `grid` (Dimension, optional) — the source viewBox unit (Radix 15, Phosphor 256, most others 24). Distinct from `size.*`, which is the CSS render size.
+* `size` (map\<string, Dimension>, optional) — named render sizes.
+* `color` (Color or token reference, optional) — default icon color; inherits via CSS `currentColor` unless explicitly set.
+
 ## Components
 
 This section provides style guidance for component atoms within the design system. The following are common component types. Design systems are encouraged to define additional components relevant to their domain.
@@ -341,13 +386,17 @@ The following names are commonly used across design systems. They are not requir
 
 **Rounded:** `none`, `sm`, `md`, `lg`, `xl`, `full`
 
+**Icon-styles:** `outlined`, `filled`, `rounded`, `sharp`, `duotone`
+
+**Icon-sizes:** `sm`, `md`, `lg`
+
 # Consumer Behavior for Unknown Content
 
 When a DESIGN.md consumer encounters content not defined by this spec:
 
 | Scenario | Behavior | Example |
 |---|---|---|
-| Unknown section heading | Preserve; do not error | `## Iconography` |
+| Unknown section heading | Preserve; do not error | `## Illustrations` |
 | Unknown color token name | Accept if value is valid | `surface-container-high: '#ede7dd'` |
 | Unknown typography token name | Accept as valid typography | `telemetry-data` |
 | Unknown spacing value | Accept; store as string if not a valid dimension | `grid-columns: '5'` |
--- a/examples/paws-and-paths/DESIGN.md
+++ b/examples/paws-and-paths/DESIGN.md
@@ -154,6 +154,16 @@ components:
     typography: "{typography.label-sm}"
     rounded: "{rounded.full}"
     padding: "{spacing.xs}"
+icons:
+  library: "Lucide"
+  style: "outlined"
+  strokeWidth: 1.75
+  grid: "24px"
+  size:
+    sm: "16px"
+    md: "20px"
+    lg: "24px"
+  color: "{colors.on-surface}"
 ---
 
 ## Brand & Style
@@ -202,7 +212,14 @@ The shape language is defined by **Rounded** corners, mirroring the soft feature
 - **Buttons:** Main CTA buttons use a `12px` (rounded-lg) radius to feel substantial and clickable.
 - **Cards:** Dog profiles and walker cards use a `1.5rem` (rounded-xl) radius to create a soft, containerized look.
 - **Inputs:** Form fields use a `0.5rem` radius to maintain a professional yet modern appearance.
-- **Icons:** Icons should feature rounded caps and corners to harmonize with the UI's structural elements.
+
+## Iconography
+
+Icons use **Lucide** in its outlined style with a slightly heavier **1.75 stroke weight**, whose rounded line caps mirror the soft corners of the UI and reinforce the friendly, human-centric feel of the brand. Icons render on a 24px grid and inherit `on-surface` color by default so they sit comfortably against any surface tint.
+
+- **Small (16px):** Inline affordances in dense lists and metadata (e.g., status dots next to walker names).
+- **Medium (20px):** Default for buttons, form fields, and navigation — large enough to read at a glance without overpowering the label.
+- **Large (24px):** Feature tiles, empty states, and hero card actions where the icon carries meaning on its own.
 
 ## Components
 
--- a/packages/cli/src/commands/spec.test.ts
+++ b/packages/cli/src/commands/spec.test.ts
@@ -89,6 +89,6 @@ describe('spec command', () => {
     const output = JSON.parse(outputStr);
     expect(output.spec).toBeDefined();
     expect(output.rules).toBeDefined();
-    expect(output.rules.length).toBe(8);
+    expect(output.rules.length).toBe(9);
   });
 });
--- a/packages/cli/src/linter/fixture.test.ts
+++ b/packages/cli/src/linter/fixture.test.ts
@@ -58,4 +58,52 @@ describe('Fixture Test', () => {
     // We expect at least the summary info
     expect(result.summary.infos).toBeGreaterThan(0);
   });
+
+  it('lints ICONS_VALID.md with no errors related to icons', () => {
+    const path = join(import.meta.dir, 'fixtures', 'ICONS_VALID.md');
+    const content = readFileSync(path, 'utf-8');
+
+    const result = lint(content);
+
+    expect(result.designSystem.icons).toBeDefined();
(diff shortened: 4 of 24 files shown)


zagi:
Hi @BenWertoski 
i did some changes to my PR, take a look when you have a free time :)
Thanks for pointing the problem out 👍 

Cross-linking from #40 — I've extended that branch (`feat/reverse_import_with_auto_detection`) so import is *not* a no-op for icons anymore: it discovers iconography metadata from `package.json` dependencies, `--icon-*` CSS custom properties, and a top-level `icons` block in DTCG token files, then emits YAML frontmatter and a `## Iconography` body section matching this PR's `IconsBlock` contract.

**Coordination notes:**

- The two PRs are independent — neither blocks the other. #40's emission is forward-compatible with the schema proposed here, and the main-branch linter today silently ignores both the unknown `icons:` frontmatter key (parser whitelists 7 keys) and the `## Iconography` section (section-order rule filters unknowns). Whichever PR merges first, the other still works.
- Once **this** PR (#44) lands, the importer's emission becomes actively validated rather than silently tolerated. No code change needed in #40 for that to happen — the YAML shape already matches.
- One small post-merge follow-up is planned: populating `icons.size.*`, `icons.grid`, and `icons.color` into the linter's `symbolTable`. That requires the widened `ResolvedValue` union that this PR introduces, so it sits as a separate PR after both land.

**Behaviour preserved:**

- `export tailwind` and `export dtcg` continue to drop icon tokens, matching the "neither format carries icon metadata" framing in the PR description.
- Existing DESIGN.md files without an `icons:` block or `## Iconography` section parse and lint identically (additive only).

dovas-net:
not too sure about that one chef, get a contributor to look over. thanks

QuantumSorcerer02:
The latest comment from **BenWertoski** introduces a pivot in the workflow. While the PR originally focused on a spec-only addition for iconography, the interaction with **zagi** has introduced a technical nuance that requires a "contributor's eye"—specifically yours, as you were tagged directly in the thread.
Here is the derivation of the current situation and the specific points of friction to look for:
### 1. The Core Hesitation: Spec-First vs. Auto-Detection
Ben’s hesitation ("not too sure about that one chef") is likely directed at the **Reverse Import** logic introduced by zagi.
 * **Ben’s Original Scope:** The PR was intended to be "additive-only" and "spec-only." He explicitly stated that export/import would be a no-op because standard formats (Tailwind/DTCG) don't carry icon metadata.
 * **The Change:** zagi attempted to "solve" this by adding automated discovery (scanning package.json, CSS custom properties, and DTCG blocks).
### 2. Derived Technical Risks
As the contributor asked to review, the "uncertainty" Ben is feeling likely stems from these areas:
 * **Heuristics Reliability:** The importer is now guessing metadata based on --icon-* CSS naming conventions. This introduces a "magic" layer that might be too opinionated or fragile for a foundational spec.
 * **Coupling:** By making the importer functional for icons now, the two PRs (#40 and #44) become functionally linked. Ben originally wanted them to be independent.
 * **Validation Complexity:** If the importer emits YAML frontmatter based on auto-detection, the linter will now actively validate it. If the auto-detection makes a mistake, the user gets linting errors for code they didn't manually write.
### 3. Immediate Review Focus
Since you've been called in to "look over" the changes, the review should prioritize:
 * **The IconsBlock Contract:** Does the YAML shape emitted by the importer perfectly match the schema defined in spec-config.yaml?
 * **The "No-Op" Philosophy:** Should the project stick to Ben’s original plan of icons being spec-only for v1, or is zagi’s auto-detection robust enough to include now?
 * **Forward Compatibility:** Ensure that if the spec changes (e.g., adding secondaryColor), the current auto-detection logic won't break or produce invalid DESIGN.md files.
### Summary of State
| Aspect | Status |
|---|---|
| **Spec Addition** | Stable (Iconography section & tokens) |
| **Linter Logic** | Ready (validating enums and dimensions) |
| **Import/Export** | **Contested** (Automated detection vs. manual spec) |
| **Action Needed** | Validation of zagi’s reverse-import logic within the context of the design.md spec. |
Ben is essentially asking for a sanity check on whether the automated discovery of icon metadata is a "bridge too far" for this specific iteration of the spec.

dovas-net:
@QuantumSorcerer02 not too sure what this adds to the conversation, but, ok

zagi:
@BenWertoski fair enough, which bit raised your concern? just wondering what can eventually be done better...

davideast:
Hey @dovas-net, @AndrewOCC, apologies for taking so long.

There's been a lot of proposals to add specific tokens to the spec, such as #47, #49, #65. I took some time to think about the larger vision and not to hastily add to the spec.

I wrote up the full thinking in [PHILOSOPHY.md](./PHILOSOPHY.md). The short version is that the spec stays focused on a small set of universal categories (colors, typography, spacing, rounded, components) as reference points, and the format is extensible for everything else. You can add an `icons:` block to your DESIGN.md today and the linter will accept it.

I do think Iconography is a good candidate for a section in the spec. It's a universal enough category that nearly every design system has icon conventions worth documenting. So I'd like to move forward with adding an `## Iconography` section to the spec. I want to be careful about prescribing specific token requirements. The tokens in the section should be non-normative, there as example references rather than required fields. The prose is where the real value lives:

> ## Iconography
> 
> Icons use **Lucide** in the outlined style at 1.5px stroke weight.
> The icon grid is 24px with three scale stops: 16px for inline
> hints, 24px for standard UI, and 32px for empty states and
> feature callouts.
> 
> Icon color follows the surrounding text color. Never tint icons
> independently of their label. The one exception is status
> indicators (error, success, warning), which inherit from the
> corresponding semantic color token.
> 
> Do not mix icon libraries. If Lucide doesn't have what you need,
> draw a custom icon on the same 24px grid with the same 1.5px
> stroke rather than pulling from a second set.

A model reading that knows the library, the style, the stroke weight, the sizes, the color rule, and the prohibition against mixing sets. It also knows why the color follows surrounding text, which is something a `color: "{colors.on-surface}"` token can't carry on its own.

@AndrewOCC, your point about end-cap and line-join style is interesting. Those attributes matter most for icon generation, and a sentence like "Strokes use round end caps and round joins to match the soft, approachable feel of the brand" gives a model everything it needs for that decision.

I'm going to close this PR because I want to start a larger spec discussion in an issue. I really appreciate the time taken with this contribution and I hope that you'll join us in the spec discussion so we can nail down the details.

PR #51, opened 2026-04-25
fix(diff): include components in token diff and remove dead import

### LINKED ISSUE #50: bug: `diff` command silently omits components from token comparison
**Summary**

The `diff` command compares `colors`, `typography`, `rounded`, and `spacing` tokens between two `DESIGN.md` files, but completely skips `components`. Any change to a component — renaming, adding, removing, or modifying a property like `backgroundColor` or `padding` — produces no output in the diff.

---

**Steps to Reproduce**

Create two `DESIGN.md` files:

`before.md`
```yaml
---
colors:
  primary: "#1A1C1E"
components:
  button-primary:
    backgroundColor: "{colors.primary}"
    padding: 12px
---
```

`after.md`
```yaml
---
colors:
  primary: "#1A1C1E"
components:
  button-primary:
    backgroundColor: "{colors.primary}"
    padding: 16px   # changed
---
```

Run:
```bash
npx @google/design.md diff before.md after.md
```

**Expected output:** `tokens.components` showing `button-primary` as modified.

**Actual output:** `tokens.components` key is absent entirely — the padding change is invisible.

---

**Root Cause**

In `packages/cli/src/commands/diff.ts`, the `tokens` object is built by calling `diffMaps` on each token group. The `components` map on `DesignSystemState` is never included. There is also an unused `serializeDesignSystem` import in the same file.

---

**Additional Context**

`diffMaps` already handles the comparison logic correctly — `components` just needs to be serialized (since `ComponentDef` uses a `Map` internally) before being passed in.

Saatvik-GT:
Fixes #50 

The `diff` command compared colors, typography, rounded, and spacing tokens
between two DESIGN.md files, but silently skipped components entirely. Any
change to a component — added, removed, or modified properties like
`backgroundColor` or `padding` — produced no output in the diff. There was
also a dead `serializeDesignSystem` import that was never called.

## Root Cause

In `commands/diff.ts`, the `tokens` object is built by passing each
`DesignSystemState` map through `diffMaps`. The `components` map was simply
never included. `ComponentDef` stores its properties in a nested `Map`, which
`JSON.stringify` serializes as `{}`, so it could not be passed to `diffMaps`
directly without first being flattened.

## Fix

Added a `serializeComponents` helper that converts `Map<string, ComponentDef>`
into `Map<string, Record<string, unknown>>` by flattening each component's
properties via `Object.fromEntries`. The result is passed to the existing
`diffMaps` utility, consistent with how all other token groups are diffed.
Also removed the unused `serializeDesignSystem` import.

## Tests

Added `diff.test.ts` (previously did not exist) with 4 cases:

- Identical files produce no component changes
- Added component is reported under `added`
- Removed component is reported under `removed`
- Modified component property is reported under `modified`

All 207 tests pass.

### DIFF
--- a/packages/cli/src/commands/diff.test.ts
+++ b/packages/cli/src/commands/diff.test.ts
@@ -1,0 +1,99 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { describe, it, expect } from 'bun:test';
+import { lint } from '../linter/index.js';
+import { diffMaps } from '../utils.js';
+import type { ComponentDef } from '../linter/model/spec.js';
+
+function serializeComponents(components: Map<string, ComponentDef>): Map<string, Record<string, unknown>> {
+  const result = new Map<string, Record<string, unknown>>();
+  for (const [name, comp] of components) {
+    result.set(name, Object.fromEntries(comp.properties));
+  }
+  return result;
+}
+
+const BASE = `---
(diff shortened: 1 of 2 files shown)


davideast (approved):

davideast:
Thank you so much @Saatvik-GT!

Saatvik-GT:
> Thank you so much @Saatvik-GT!

Glad it landed! Happy to contribute more :)

PR #58, opened 2026-04-28
chore: update checkout action version from v4 to v6

coliff:
v4 is unsupported and v6 is safer.
https://github.com/actions/checkout/releases


davideast:
Thank you @coliff!

davideast (approved):

PR #59, opened 2026-04-28
fix(linter): orphaned-tokens should not flag MD3 paired tokens or baseline families

### LINKED ISSUE #46: bug: orphaned-tokens rule flags 33-43 false positives on each shipped example
## Summary

The `orphaned-tokens` rule fires on **every MD3 semantic color token** (`surface-*`, `on-*`, `inverse-*`, `*-container`, `*-fixed`, etc.) because they aren't referenced by any entry in `components:`. Result: every one of the three shipped examples (`examples/atmospheric-glass`, `examples/paws-and-paths`, `examples/totality-festival`) fails linting with 33–43 orphan warnings each, **zero errors**.

## Reproduce

```bash
git clone https://github.com/google-labs-code/design.md && cd design.md
bun install && bun run build   # or: cd packages/cli && npm install && npm run build
for ex in atmospheric-glass paws-and-paths totality-festival; do
  echo "=== $ex ==="
  node packages/cli/dist/index.js lint examples/$ex/DESIGN.md | jq '.summary'
done
```

Expected output:

```
=== atmospheric-glass ===
{ "errors": 0, "warnings": 43, "infos": 1 }
=== paws-and-paths ===
{ "errors": 0, "warnings": 33, "infos": 1 }
=== totality-festival ===
{ "errors": 0, "warnings": 38, "infos": 1 }
```

All warnings are of the form `'surface-container-low' is defined but never referenced by any component.`

## Why this is a false positive

MD3-style token systems are **semantic layers**, not component property primitives. Tokens like `surface-container-low`, `on-secondary-container`, or `inverse-primary` are meant to be consumed by the theme / runtime / CSS custom properties — **not declared explicitly in each component entry**. The spec itself acknowledges this non-component-driven usage:

> A common convention is to name the palettes in this order: primary, secondary, tertiary, and neutral. (spec.md § Colors)

The rule assumes every color token must have a `components:` consumer, which inverts the semantic-layer pattern the examples themselves model.

## Proposed fix

Skip the rule when the token name matches the MD3 semantic namespace:

```
^(surface|on-|inverse-|primary-|secondary-|tertiary-|neutral-|error-|background|scrim|shadow)
```

Or, more conservatively, downgrade `orphaned-tokens` to `severity: info` for tokens matching that pattern. Authors who genuinely want strict orphan detection can re-raise severity via a custom rule.

Happy to open a PR if this direction is acceptable — have a working patch against `linter/rules/orphaned-tokens.ts` ready to submit once the CLA is signed.

## Context

Caught while integrating `@google/design.md` as a library in an external bridge (`oknemix/os`, `tools/design-md-bridge`). All 299 of our regression tests pass, but the orphaned-tokens signal was too noisy to surface real issues until we mapped the pattern. Sharing upstream because every user of MD3-shaped tokens hits the same wall.

mvanhorn:
## Summary

Makes the `orphaned-tokens` lint rule aware of MD3 token families and the MD3 baseline contract, so the three shipped examples stop emitting 33-43 false-positive warnings each.

## Why this matters

Per [#46](https://github.com/google-labs-code/design.md/issues/46):

> The `orphaned-tokens` rule fires on **every MD3 semantic color token** (`surface-*`, `on-*`, `inverse-*`, `*-container`, `*-fixed`, etc.) because they aren't referenced by any entry in `components:`. Result: every one of the three shipped examples (`examples/atmospheric-glass`, `examples/paws-and-paths`, `examples/totality-festival`) reports 33-43 false orphans.

Verified on `main`:

| Example | Orphan findings before | After this PR |
|---|---:|---:|
| `atmospheric-glass` | 43 | 0 |
| `paws-and-paths` | 33 | 0 |
| `totality-festival` | 38 | 0 |

The example DESIGN.md files declare the full MD3 baseline palette but the canonical components only directly reference a small slice (`primary`, `on-primary`, `primary-fixed-dim`, `on-surface-variant`). Tokens like `on-primary`, `primary-container`, `inverse-primary` aren't decoration — they're part of the same semantic family that consumers use at runtime. The rule was treating "no component direct reference" as "orphaned", which produces noise on every shipped example.

## Changes

`packages/cli/src/linter/linter/rules/orphaned-tokens.ts`:

1. New `colorFamily(name)` helper that reduces an MD3 token name to its family root by stripping prefixes (`on-`, `inverse-`) and suffixes (`-container*`, `-fixed*`, `-dim`, `-bright`, `-tint`, `-variant`). Custom tokens that don't match the patterns collapse to themselves, so `brand-blue` keeps getting flagged when unused.
2. **Family-aware reference tracking**: when a component references one token (say `colors.primary`), every sibling in the same family (`on-primary`, `primary-container`, `on-primary-container`, `primary-fixed`, `primary-fixed-dim`, `on-primary-fixed`, `on-primary-fixed-variant`, `inverse-primary`) is treated as part of the in-use group.
3. **MD3 baseline families exempt**: the standard MD3 set (primary, secondary, tertiary, error, surface, background, outline) is part of the contract a design system ships, not optional decoration. Tokens belonging to those families don't get flagged even when no component references them yet.

`packages/cli/src/linter/linter/rules/orphaned-tokens.test.ts`:
- Adds 4 new test cases covering: MD3 paired tokens via referenced family, surface family with sibling reference, custom tokens still getting flagged, MD3 baseline families exempt without any reference.

## Testing

- `bun test src/linter/linter/rules/orphaned-tokens.test.ts` — 6 pass, 0 fail.
- `bun test` (full suite) — 207 pass, 0 fail (was 206 before; +1 net new test that consolidated; full run added 4 new tests as above).
- `bun run src/index.ts lint examples/<each>/DESIGN.md` confirms the orphan count drops to 0 on all three shipped examples.

The rule still flags genuinely-orphaned custom tokens (regression test asserts `brand-blue` still fires when defined but unreferenced).

Fixes #46

This contribution was developed with AI assistance.

### DIFF
--- a/packages/cli/src/linter/linter/rules/orphaned-tokens.test.ts
+++ b/packages/cli/src/linter/linter/rules/orphaned-tokens.test.ts
@@ -33,4 +33,103 @@ describe('orphanedTokens', () => {
     const state = buildState({ colors: { primary: '#ff0000' } });
     expect(orphanedTokens(state)).toEqual([]);
   });
+
+  it('does not flag MD3 paired tokens when the family is referenced (issue #46)', () => {
+    // When a component references `primary`, the rest of the MD3 primary
+    // family (`on-primary`, `primary-container`, `on-primary-container`,
+    // `primary-fixed`, `primary-fixed-dim`, `on-primary-fixed`,
+    // `on-primary-fixed-variant`, `inverse-primary`) is part of the same
+    // semantic group and should not be flagged as orphaned.
+    const state = buildState({
+      colors: {
+        primary: '#1A1C1E',
+        'on-primary': '#ffffff',
+        'primary-container': '#e2e2e2',
+        'on-primary-container': '#636565',
+        'primary-fixed': '#e2e2e2',
+        'primary-fixed-dim': '#c6c6c7',
+        'on-primary-fixed': '#1a1c1c',
+        'on-primary-fixed-variant': '#454747',
+        'inverse-primary': '#5d5f5f',
+      },
+      components: {
+        button: { backgroundColor: '{colors.primary}' },
+      },
+    });
+    const findings = orphanedTokens(state);
+    expect(findings).toEqual([]);
+  });
+
+  it('does not flag MD3 surface family when one surface token is referenced', () => {
+    const state = buildState({
+      colors: {
+        surface: '#0b1326',
+        'surface-dim': '#0b1326',
+        'surface-bright': '#31394d',
+        'surface-container': '#171f33',
+        'surface-container-lowest': '#060e20',
+        'surface-container-low': '#131b2e',
+        'surface-container-high': '#222a3d',
+        'surface-container-highest': '#2d3449',
+        'on-surface': '#dae2fd',
+        'on-surface-variant': '#c4c7c8',
+        'inverse-surface': '#dae2fd',
+        'inverse-on-surface': '#283044',
+        'surface-tint': '#c6c6c7',
+        'surface-variant': '#2d3449',
+      },
+      components: {
+        card: { backgroundColor: '{colors.surface-container}' },
+      },
+    });
+    const findings = orphanedTokens(state);
+    expect(findings).toEqual([]);
+  });
+
+  it('still flags genuinely-orphaned custom tokens outside any referenced family', () => {
+    // `brand-blue` is not part of the MD3 primary family, so referencing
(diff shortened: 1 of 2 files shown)


davideast:
Thank you @mvanhorn! I'll get this merged in real quick to silence the annoying errors. This has me thinking about providing different configurations for linting depending on a specific system being followed. This would be opt-in and extensible. Let me know if you have any thoughts.

PR #67, opened 2026-05-03
feat(typography): add base font-size and paragraph-max-width tokens

### LINKED ISSUE #65: Base font size and paragraph max-width
It would be nice for the spec to provide for…

1. base font-size
2. heading and paragraph max-width tokens. I usually define them for headings, body text and lead text. 

Maybe these could be included under typography, like the following?

``` yaml
typography:
  base: 16px
  measure:
    heading: 28rem
    lead: 36rem
    body: 38rem
```

Or, maybe somewhere else?

Angfr95:
Fixes #65

### Context

`design.md` is a CLI that parses DESIGN.md files (YAML design tokens) and exports them to various formats (Tailwind v4 CSS, JSON, DTCG). The existing `typography` section only supported composite tokens (objects with `fontFamily`, `fontSize`, `fontWeight`, etc.) — there was no place for a scalar font-size value or a `paragraph-max-width` group.

### What this PR adds

Two new top-level token sections:

- `font-size` → emits `--typography-{name}` CSS custom properties
- `paragraph-max-width` → emits `--paragraph-max-width-{name}` CSS custom properties

> Implemented as top-level sections (consistent with `spacing`, `colors`) rather than nested under `typography`, since the existing typography tokens are composite objects — mixing scalar and composite tokens in the same section would complicate the parsing logic.

### Usage

```yaml
font-size:
  base: 16px

paragraph-max-width:
  lead: 46rem
  body: 37rem
```

Emits:

```css
@theme {
  --typography-base: 16px;
  --paragraph-max-width-lead: 46rem;
  --paragraph-max-width-body: 37rem;
}
```

### Files changed

| File | Change |
|------|--------|
| `spec-config.yaml` | Added examples + recommended tokens for both sections |
| `spec-config.ts` | Extended Zod schema to validate new fields |
| `parser/spec.ts` | Added `fontSizeScale` and `paragraphMaxWidth` to `ParsedDesignSystem` |
| `parser/handler.ts` | Mapped YAML keys `font-size` / `paragraph-max-width` to TS fields |
| `model/spec.ts` | Added two new `Map<string, ResolvedDimension>` to `DesignSystemState` |
| `model/handler.ts` | Parsing + validation logic (same pattern as `spacing`) |
| `tailwind/v4/spec.ts` | Two new optional fields on `TailwindV4ThemeData` |
| `tailwind/v4/handler.ts` | Emission of both new categories |
| `tailwind/v4/serialize.ts` | CSS serialization with `--typography-` and `--paragraph-max-width-` prefixes |
| `spec-gen/` | Docs, renderers, compiler scope fix |
| `docs/spec.md` | Regenerated via `bun run spec:gen` |

**232 tests, 0 failures.**


michaelGregoire:
After some thinking about this, it might be wise to use "measure", the traditional typographic term for this concept, rather than paragraph-max-width.

davideast:
Hey @Angfr95, thank you for this. The implementation is thorough and it's clear you put real work into it.

Since this was filed we published a [PHILOSOPHY.md](./PHILOSOPHY.md) that explains the direction for how the spec evolves. The core principle is that the spec defines a structural minimum. The prose sections are where design intent lives, and the model already understands CSS. It doesn't need `font-size` or `paragraph-max-width` as dedicated top-level sections to interpret those concepts from a DESIGN.md.

Adding new top-level token categories is something we're being careful about because there's no natural stopping point. `font-size` and `paragraph-max-width` are valid, but so are `line-height`, `letter-spacing`, `text-indent`, `word-spacing`, and dozens of other properties. Each one is a reasonable individual request, but together they push the spec toward being a CSS property registry, which is not the goal.

@michaelGregoire (who filed #65), your suggestion to nest these under `typography` is closer to where we'd go if we expanded. But we think the better path is a configurable linter. We're working on letting projects register custom token sections and properties so the linter recognizes them without spec changes. That way your project can use `font-size` and `measure` (good call on the typographic term) as first-class tokens without every DESIGN.md in the world needing to support them.

Going to close this for now, but the use case from #65 is real and will be addressable through configuration. Thank you both.

PR #72, opened 2026-05-05
fix: allow Tailwind v4 digit-leading token names

### LINKED ISSUE #71: Tailwind v4 `css-tailwind` export rejects digit-prefix token names that Tailwind v4 itself uses
## Summary

The `css-tailwind` exporter (Tailwind v4 `@theme {}`) validates token names against `^[a-zA-Z][a-zA-Z0-9-]*$`, which rejects names starting with a digit. This is more restrictive than CSS spec actually requires for custom properties, AND it rejects names that Tailwind v4 itself uses by default (`--radius-2xl`, `--radius-3xl`, `--spacing-2`, etc.).

## Reproducer

```yaml
---
name: Repro
colors:
  primary: "#ED5A2F"
spacing:
  2xs: 2px
  xs: 4px
---

## Overview

Minimal repro.
```

```sh
$ npx --yes @google/design.md export --format css-tailwind DESIGN.md
{"error":"Token name \"2xs\" is not a valid CSS identifier for Tailwind v4 export (must match /^[a-zA-Z][a-zA-Z0-9-]*$/)."}
```

The same input lints clean and exports fine via `--format tailwind` (Tailwind v3 JSON) and `--format dtcg`.

## Why this should work

1. **CSS spec allows it.** Per [css-variables-1 §2](https://www.w3.org/TR/css-variables-1/#defining-variables), custom property names are `--<ident>`. CSS identifiers ([css-syntax-3 §4.1.1](https://www.w3.org/TR/css-syntax-3/#typedef-ident-token)) allow digits anywhere except as the literal first character of an unprefixed ident — but the `--` prefix means the post-prefix part is treated as `<custom-property-name>`, which permits digit-leading content.

2. **Tailwind v4 itself uses these.** Tailwind v4's default theme defines `--radius-2xl`, `--radius-3xl`, `--radius-4xl`, `--spacing-0`, `--spacing-0\.5`, `--spacing-2`, etc. A tool that emits Tailwind v4 `@theme {}` shouldn't enforce a stricter naming rule than Tailwind v4's own conventions.

3. **The other emitters (`tailwind`, `dtcg`) accept the same names.** It's only the v4 path that blocks.

## Source

`packages/cli/src/linter/tailwind/v4/handler.ts:42`:

```ts
const VALID_TOKEN_NAME = /^[a-zA-Z][a-zA-Z0-9-]*$/;
```

## Proposed fix

```diff
-const VALID_TOKEN_NAME = /^[a-zA-Z][a-zA-Z0-9-]*$/;
+const VALID_TOKEN_NAME = /^[a-zA-Z0-9][a-zA-Z0-9-]*$/;
```

(Or relax further — allow `_` and dot-escape — if you want full Tailwind v4 parity for fractional spacing.)

Happy to send a PR if the proposed fix is the direction you want.

## Context

Discovered while wiring `@google/design.md` into a multi-tier scale (`2xs/xs/sm/md/lg/xl/2xl/3xl/4xl/5xl/6xl`) that mirrors Tailwind v4's defaults. The lint passes and `tailwind`/`dtcg` exports both work; only `css-tailwind` blocks, which is the one we actually need for a Tailwind v4 build.

Bortlesboat:
## Summary
- relax the Tailwind v4 token-name validator so names like \\2xs\\ and \\4xl\\ can export
- keep rejecting dotted and space-containing token names
- update the Tailwind v4 emitter regression test for digit-leading spacing and radius tokens

Fixes #71.

## Verification
- RED: \\
px bun test packages/cli/src/linter/tailwind/v4/handler.test.ts\\ failed on the new digit-leading token test before the handler change
- GREEN: \\
px bun test packages/cli/src/linter/tailwind/v4/handler.test.ts\\ -> 9 pass, 0 fail
- \\
px bun test packages/cli/src/linter/tailwind/v4\\ -> 17 pass, 0 fail
- \\git diff --check\\

## Note
- \\./node_modules/.bin/tsc.exe --project packages/cli/tsconfig.build.json --noEmit\\ is blocked locally by dependency declaration errors in \\un-types\\ under the transient \\
px bun\\ install before reaching this change.

### DIFF
--- a/packages/cli/src/linter/tailwind/v4/handler.test.ts
+++ b/packages/cli/src/linter/tailwind/v4/handler.test.ts
@@ -144,13 +144,15 @@ describe('TailwindV4EmitterHandler', () => {
       expect(result.error.message).toContain('primary.surface');
     });
 
-    it('fails when a spacing token name starts with a digit', () => {
+    it('allows Tailwind-style dimension token names that start with a digit', () => {
       const state = buildState({});
-      state.spacing.set('1bad', { type: 'dimension', value: 4, unit: 'px' });
+      state.spacing.set('2xs', { type: 'dimension', value: 2, unit: 'px' });
+      state.rounded.set('4xl', { type: 'dimension', value: 32, unit: 'px' });
       const result = emitter.execute(state);
-      expect(result.success).toBe(false);
-      if (result.success) return;
-      expect(result.error.code).toBe('INVALID_TOKEN_NAME');
+      expect(result.success).toBe(true);
+      if (!result.success) return;
+      expect(result.data.theme.spacing?.['2xs']).toBe('2px');
+      expect(result.data.theme.borderRadius?.['4xl']).toBe('32px');
     });
 
     it('fails when a token name contains a space', () => {
--- a/packages/cli/src/linter/tailwind/v4/handler.ts
+++ b/packages/cli/src/linter/tailwind/v4/handler.ts
@@ -15,7 +15,7 @@
 import type { TailwindV4EmitterSpec, TailwindV4EmitterResult, TailwindV4ThemeData } from './spec.js';
 import type { DesignSystemState, ResolvedDimension } from '../../model/spec.js';
 
-const VALID_TOKEN_NAME = /^[a-zA-Z][a-zA-Z0-9-]*$/;
+const VALID_TOKEN_NAME = /^[a-zA-Z0-9][a-zA-Z0-9-]*$/;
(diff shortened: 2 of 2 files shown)


davideast:
Hey @Bortlesboat! Apologies on the delay. I'm going to give this a review here soon.

davideast:
@Bortlesboat Thank you so much. Looks good. Clean fix, good test update.

davideast (approved):

PR #74, opened 2026-05-06
feat: add motion tokens (durations, easings, transitions)

sbrsubuvga:
Adds a  token group to DESIGN.md alongside ,
, , and . Three sub-groups:

-  — ,  (ms or s)
-    — cubic-bezier control points
-  — composite of

Components consume transitions via a single new sub-token, , that takes a token reference. Asymmetric in/out
timing is expressed via component variants (e.g. , ) — same pattern as .

Cross-platform by design: easings are stored as control-point arrays (the mathematical contract every animation engine implements), not CSS strings. Exports map cleanly to:

- Tailwind v4 — , ,
- Tailwind v3 — ,
- DTCG       —  /  /
                (W3C Design Tokens Format Module 2025.10)

Motion is fully optional. DESIGN.md files without  lint, export, and diff identically to before this change.

Includes:
- New canonical section  (between Shapes and Components)
- 12 new unit tests (model handler, Tailwind v4, DTCG)
- Regenerated docs/spec.md with motion schema, types, and section
- New example: examples/heritage/ demonstrating motion in context

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,6 +113,15 @@ rounded:
   <scale-level>: <Dimension>
 spacing:
   <scale-level>: <Dimension | number>
+motion:
+  duration:
+    <token-name>: <Duration>
+  easing:
+    <token-name>: <CubicBezier>
+  transition:
+    <token-name>:
+      duration: <Duration | token reference>
+      easing: <CubicBezier | token reference>
 components:
   <component-name>:
     <token-name>: <string | token reference>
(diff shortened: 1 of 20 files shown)


davideast:
Hey @sbrsubuvga, thank you for this. I appreciate that this was a lot of work with 983 lines across 20 files with tests, a new example, and regenerated docs.

Since this was filed we published a [PHILOSOPHY.md](./PHILOSOPHY.md) that lays out the direction for the spec. The core principle is that the spec defines a structural minimum. The prose sections are where design intent lives. An LLM reading a DESIGN.md already understands motion, easing curves, and transitions from the prose. It doesn't need `durations`, `easings`, and `transitions` as structured token categories to work with those concepts.

Motion is one of several domain-specific token categories that have been proposed (iconography, elevation, text-transform, font-size scales). Each one is individually reasonable, but together they push the spec toward being a comprehensive design token registry. That's not the goal. The spec stays small so that the prose can do the expressive work.

What we're building instead is a configurable linter that lets projects register custom token categories and properties. Your project could define `motion` as a recognized section with `durations`, `easings`, and `transitions` sub-tokens, and the linter would validate them without needing any spec change. That gives you structured motion tokens for your system without requiring every DESIGN.md to support them.

Going to close this for now. The use case is valid and the implementation quality is high. When linter configuration lands, motion support like this could live as a shareable rule pack that projects opt into. Thank you for the contribution.

PR #84, opened 2026-05-18
feat(linter): add unknown-key rule for unknown top-level keys

### LINKED ISSUE #83: Feature: warn on unknown top-level schema keys (catch typos like `colours:`)
## Problem

The parser silently discards unknown top-level YAML keys. [`ParserHandler.toDesignSystem`](https://github.com/google-labs-code/design.md/blob/887a4bff8943f37ef7c9a2d8c01266f18e3fe895/packages/cli/src/linter/parser/handler.ts#L192-L205) builds the design system from a fixed whitelist (`name`, `description`, `colors`, `typography`, `rounded`, `spacing`, `components`); any key not in that list is dropped without an error or a warning. A misspelled section name therefore disappears with no signal to the author:

```yaml
---
name: "Example"
colours:               # typo for `colors` — silently ignored
  primary: "#E30613"
---
```

`lint` reports no findings for the `colours` block, and downstream `missing-sections` even reports `colors` as absent — the author is told the *opposite* of what they intended, with no hint that a typo is the cause. There is currently no way for a user to notice this class of mistake.

## Proposal

Add a lint rule that warns when the front matter contains a top-level key that is not part of the known schema.

**Known top-level keys** (allowlist): `name`, `description`, `colors`, `typography`, `rounded`, `spacing`, `components`, `version`.

**New linting rule**

| Rule | Severity | What it checks |
|---|---|---|
| `unknown-key` | warning | A top-level front-matter key is not part of the known schema |

**Proposed message**

> `Unexpected unknown top-level key "colours"`

## Example output

```json
{
  "findings": [
    {
      "severity": "warning",
      "path": "colours",
      "message": "Unexpected unknown top-level key \"colours\""
    }
  ],
  "summary": { "errors": 0, "warnings": 1, "infos": 0 }
}
```

## Version

`@google/design.md` 0.1.1

ryo-manba:
Closes #83

Adds an `unknown-key` lint rule (warning) that flags top-level YAML keys not in the known schema, so a typo like `colours:` is reported instead of silently discarded.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ npx @google/design.md spec --rules-only --format json
 
 ## Linting Rules
 
-The linter runs seven rules against a parsed DESIGN.md. Each rule produces findings at a fixed severity level.
+The linter runs nine rules against a parsed DESIGN.md. Each rule produces findings at a fixed severity level.
 
 | Rule | Severity | What it checks |
 |:-----|:---------|:---------------|
@@ -305,6 +305,7 @@ The linter runs seven rules against a parsed DESIGN.md. Each rule produces findi
 | `missing-sections` | info | Optional sections (spacing, rounded) absent when other tokens exist |
 | `missing-typography` | warning | Colors are defined but no typography tokens exist — agents will use default fonts |
 | `section-order` | warning | Sections appear out of the canonical order defined by the spec |
+| `unknown-key` | warning | A top-level YAML key is not part of the known schema (catches typos like `colours:`) |
 
 ### Programmatic API
 
--- a/README.md
+++ b/README.md
@@ -305,7 +305,7 @@ The linter runs nine rules against a parsed DESIGN.md. Each rule produces findin
 | `missing-sections` | info | Optional sections (spacing, rounded) absent when other tokens exist |
 | `missing-typography` | warning | Colors are defined but no typography tokens exist — agents will use default fonts |
 | `section-order` | warning | Sections appear out of the canonical order defined by the spec |
-| `unknown-key` | warning | A top-level YAML key is not part of the known schema (catches typos like `colours:`) |
+| `unknown-key` | warning | A top-level YAML key looks like a typo of a known schema key (e.g. `colours:` → `colors:`); custom extension keys stay silent |
 
 ### Programmatic API
 
--- a/packages/cli/src/commands/spec.test.ts
+++ b/packages/cli/src/commands/spec.test.ts
@@ -89,6 +89,6 @@ describe('spec command', () => {
     const output = JSON.parse(outputStr);
     expect(output.spec).toBeDefined();
     expect(output.rules).toBeDefined();
-    expect(output.rules.length).toBe(8);
+    expect(output.rules.length).toBe(9);
   });
 });
--- a/packages/cli/src/linter/index.test.ts
+++ b/packages/cli/src/linter/index.test.ts
@@ -140,10 +140,27 @@ colours:
     const result = lint(content);
 
     const finding = result.findings.find(
-      f => f.message === 'Unexpected unknown top-level key "colours"'
+      f => f.message === 'Unknown key "colours" — did you mean "colors"?'
     );
     expect(finding).toBeDefined();
     expect(finding!.severity).toBe('warning');
     expect(finding!.path).toBe('colours');
+  });
+
+  it('stays silent for custom extension keys that are not close to any known key', () => {
+    const content = `---
+name: Example
+icons:
+  search: "search-icon.svg"
+motion:
+  fast: "100ms"
+---`;
+
+    const result = lint(content);
+
+    const unknownKeyFindings = result.findings.filter(f =>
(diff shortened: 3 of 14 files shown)


LealLara (approved):

davideast:
Hey @ryo-manba! Thank you so much for the issue and PR.

I agree this is useful for typos. However, we need to consider the primary case of custom token values in DESIGN.md. We don't require anything but `name`, no token sections are strictly required. 

The only one conditional requirement is in `colors`, which if present the `primary` color palette must be defined.  All major sections: `colors`, `typography`, `rounded`, `spacing`, `components` are optional. And the recommended token names (`primary`, `secondary`, `headline-lg`, `sm`, etc.) are guidance, not requirements.

So the idea here would be to identify typos for a warn rather than any unknown token values because we shouldn't trigger warn logs for custom fields.

Here's what I would explore for a typo-based warn rule.

The `ParsedDesignSystem` interface in `spec.ts` is the canonical source of known keys. So you can use an approach like an Levenshtein-based typo detection, with the known keys derived from the `ParsedDesignSystem` interface.

#### Derive keys from the schema, don't hardcode them

This PR currently hardcodes `KNOWN_TOP_LEVEL_KEYS` in `handler.ts`. Instead, derive them from the `ParsedDesignSystem` interface's property names. Since TypeScript interfaces don't exist at runtime, you need a single source-of-truth constant that both the type and the rule reference:

```typescript
// packages/cli/src/linter/parser/spec.ts

/** Canonical top-level YAML keys per the DESIGN.md schema. */
export const SCHEMA_KEYS = [
  'version', 'name', 'description',
  'colors', 'typography', 'rounded', 'spacing', 'components',
] as const;

export type SchemaKey = typeof SCHEMA_KEYS[number];
```

This gives us runtime keys that also work for type inference.

#### Add a zero-dependency Levenshtein function

This is ~15 lines, so no npm dependency needed:

```typescript
// packages/cli/src/linter/linter/rules/levenshtein.ts

/** Levenshtein edit distance between two strings. */
export function levenshtein(a: string, b: string): number {
  const m = a.length, n = b.length;
  const dp: number[][] = Array.from({ length: m + 1 }, (_, i) =>
    Array.from({ length: n + 1 }, (_, j) => (i === 0 ? j : j === 0 ? i : 0))
  );
  for (let i = 1; i <= m; i++) {
    for (let j = 1; j <= n; j++) {
      dp[i][j] = a[i - 1] === b[j - 1]
        ? dp[i - 1][j - 1]
        : 1 + Math.min(dp[i - 1][j], dp[i][j - 1], dp[i - 1][j - 1]);
    }
  }
  return dp[m][n];
}
```

#### Rewrite the rule: only warn when a close match exists

```typescript
// packages/cli/src/linter/linter/rules/unknown-key.ts

import { SCHEMA_KEYS } from '../../parser/spec.js';
import { levenshtein } from './levenshtein.js';
import type { DesignSystemState } from '../../model/spec.js';
import type { RuleDescriptor, RuleFinding } from './types.js';

/** Max edit distance to consider a typo (not a custom key). */
const MAX_TYPO_DISTANCE = 2;

export function unknownKey(state: DesignSystemState): RuleFinding[] {
  const knownSet = new Set<string>(SCHEMA_KEYS);
  return (state.unknownKeys ?? []).flatMap(key => {
    if (knownSet.has(key)) return [];
    
    // Find closest known key
    let bestMatch: string | undefined;
    let bestDist = Infinity;
    for (const known of SCHEMA_KEYS) {
      const dist = levenshtein(key.toLowerCase(), known.toLowerCase());
      if (dist < bestDist) {
        bestDist = dist;
        bestMatch = known;
      }
    }

    if (bestDist <= MAX_TYPO_DISTANCE && bestMatch) {
      return [{
        path: key,
        message: `Unknown key "${key}" — did you mean "${bestMatch}"?`,
      }];
    }
    
    // Not close to any known key → intentional extension, stay silent
    return [];
  });
}
```

#### Key heuristics
- **Threshold**: A `distance ≤ 2` catches `colours`→`colors` (1), `typografy`→`typography` (2), but ignores `icons`, `motion`, `brand.
- **Case handling**: Compare lowercased. This way `Colors` and `COLORS` should match `colors`.
- **Silence for zero matches**: Don't emit any finding to respect the spec's extensibility philosophy.

#### Test cases to cover

```
"colours"     → warns, suggests "colors"       (distance 1)
"typografy"   → warns, suggests "typography"    (distance 2)  
"icons"       → silent (distance 4 from any key)
"motion"      → silent (distance 5+ from any key)
"colors"      → silent (exact match, not unknown)
"nam"         → warns, suggests "name"          (distance 1)
"rounding"    → silent (distance 3 from "rounded")
```

davideast (changes requested):
Requesting changes based on the larger comment I left.

ryo-manba:
@davideast Thanks for the review! I've fixed it.

davideast (approved):
LGTM. Thank you @ryo-manba for addressing the feedback!

Issue #47, opened 2026-04-24
Feature: motion tokens (duration, easing) with prefers-reduced-motion fallback


oknemixam:
## Summary

DESIGN.md currently has no way to express motion — duration, easing, transitions — which means agents generating code have to invent defaults every time. Proposing a `motion:` frontmatter block and a `## Motion` markdown section.

## Precedent

- W3C DTCG 2025.10 does not cover motion tokens (confirmed by reading the current draft)
- Material Design 3 ships duration + easing tokens (`md.motion.duration.short1` = 50ms, etc.)
- IBM Carbon, Fluent UI, Shopify Polaris all publish motion tokens
- Radix UI and Tailwind ship motion utilities but no formal tokens
- Missing link: a format that lets a design system publish motion tokens as first-class entries that an agent or exporter can read

## Proposed schema

```yaml
motion:
  duration:
    fast: 150ms
    medium: 300ms
    slow: 600ms
  easing:
    standard:    "cubic-bezier(0.2, 0, 0, 1)"
    emphasized:  "cubic-bezier(0.2, 0, 0, 1.5)"
    decelerate:  "cubic-bezier(0, 0, 0, 1)"
  prefers-reduced-motion-fallback:
    medium: 100ms    # 300ms collapses to 100ms
    slow:   100ms    # 600ms collapses to 100ms
```

Parallel to existing `spacing` / `rounded` maps. Values are CSS-native (ms/s for duration, `cubic-bezier(…)` or named keyword for easing).

## Proposed lint rule

`motion-reduce-compat` — warn when a duration exceeds 200ms and no matching key exists in `prefers-reduced-motion-fallback`. Rationale: animations longer than roughly 200ms register as **deliberate motion** (rather than state transitions), which WCAG 2.2 + vestibular-sensitivity guidance says needs a reduced-motion path.

Severity: warning. Authors who intentionally ship long animations without a fallback (brand showcase, hero, etc.) acknowledge the warning explicitly.

## Implementation notes

Built and verified in our `@google/design.md` bridge:

- `motion.ts` ~100 lines — parse, validate
- `test-motion.ts` 27/27 passing
- Integrates cleanly as an additive `motion:` key (Google's parser silently drops unknown frontmatter keys, so this is backward-compatible with v0.1.1 until the spec lands)

Happy to PR either as:

1. Spec-only (just adopt the schema into `docs/spec.md`)
2. Spec + linter rule (adds the rule to `DEFAULT_RULES`)
[...]

PR #89, opened 2026-05-25
feat(spec): add 9 standard CSS component sub-tokens

tn819:
## Summary

Extends the `component_sub_tokens` vocabulary from 8 to 17 entries to support real-world design systems that use standard CSS properties within component definitions.

## Problem

The current 8-token vocabulary (`backgroundColor`, `textColor`, `typography`, `rounded`, `padding`, `size`, `height`, `width`) is insufficient for production design systems. Real-world components need additional standard CSS properties:

| Missing Token | Real-World Usage |
|---|---|
| `minHeight` / `minWidth` | Touch-target compliance (44 px, 56 px buttons) |
| `fontFamily` / `fontWeight` / `fontStyle` | Per-component type overrides (serif CTAs, italic margin-notes) |
| `borderColor` | Card, chip, and container borders |
| `borderLeftColor` / `borderLeftWidth` | Directional border patterns (left-edge accent) |
| `paddingLeft` | Directional padding (margin-note inset) |
| `elevation` | Shadow classes (floating SOS/elevated buttons) |

Without these in the spec, every real design system that uses them gets a wall of "not a recognized component sub-token" warnings. This makes `lint` output noisy and obscures actual issues.

## Change

Adds 9 new entries to `component_sub_tokens` in `spec-config.yaml`, each with:
- A descriptive name matching CSS naming conventions
- An appropriate type (`Color`, `Dimension`, `string`, or `number | string`)
- An optional description with usage context

All additions follow existing spec patterns. Directional variants (`borderLeftColor`, `paddingLeft`) use standard CSS box-model naming.

## Testing

Existing tests should pass (no token removals, only additions). The change is backwards-compatible — existing valid DESIGN.md files continue to validate, and the new tokens are recognized.

### DIFF
--- a/packages/cli/src/linter/spec-config.yaml
+++ b/packages/cli/src/linter/spec-config.yaml
@@ -70,12 +70,42 @@ component_sub_tokens:
     type: Dimension
   - name: padding
     type: Dimension
+  - name: paddingLeft
+    type: Dimension
+    description: "CSS logical directional padding, e.g. for margin-note inset patterns."
   - name: size
     type: Dimension
   - name: height
     type: Dimension
   - name: width
     type: Dimension
+  - name: minHeight
+    type: Dimension
+    description: "Minimum height (commonly 44px for touch-target compliance)."
+  - name: minWidth
+    type: Dimension
+    description: "Minimum width (e.g. circular buttons, touch-target compliance)."
+  - name: fontFamily
+    type: string
+    description: "Font family override per component (e.g. serif for CTAs, sans for chips)."
+  - name: fontWeight
+    type: "number | string"
+    description: "Font weight override per component."
+  - name: fontStyle
+    type: string
+    description: "CSS font-style value (e.g. 'italic' for margin-note patterns)."
+  - name: borderColor
+    type: Color
+    description: "Border color for card, chip, and container edges."
+  - name: borderLeftColor
+    type: Color
+    description: "Directional border color (e.g. accent left-edge for margin-notes)."
+  - name: borderLeftWidth
+    type: Dimension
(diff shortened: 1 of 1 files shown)


davideast:
Hey @tn819, thank you for this. The table of real-world usage is helpful and clearly thought through.

I've been working through a set of similar proposals (#41, #47, #49, #65) and recently published a [PHILOSOPHY.md](./PHILOSOPHY.md) that explains the direction for how the spec evolves. The spec defines a structural minimum and the prose is where the design lives. We're cautious about expanding the token vocabulary toward CSS coverage because the model already knows CSS. It doesn't need `minHeight` or `borderLeftColor` re-encoded in the spec to understand those properties.

That said, the problem you're describing is real. Getting a wall of "not a recognized component sub-token" warnings for standard properties like `fontFamily` or `borderColor` is noisy and makes the linter output less useful.

I think the fix is in the warning behavior, not in the token list. Right now the linter warns on every unknown component sub-token. That made sense when the list was meant to be exhaustive, but it doesn't fit a format that's designed to be extensible. A better approach would be to either:

1. **Downgrade the severity.** Unknown sub-tokens could be an `info` instead of a `warning`, so they show up but don't clutter the output alongside actual problems.
2. **Only warn on likely typos.** Similar to how the `unknown-key` rule works at the top level — use Levenshtein distance to warn when a sub-token looks like a misspelling of a recognized one, and stay silent for clearly intentional custom properties.

Either approach solves the noisy-output problem without adding CSS properties to the spec one by one, which is a path that doesn't have a natural stopping point.

Would either of those approaches address what you're running into? If you're interested in taking on the fix, I'd be happy to work with you on it.

PR #105, opened 2026-06-13
feat: warn on token-like frontmatter keys silently ignored by export

### LINKED ISSUE #100: lint should warn when frontmatter token-like fields are ignored by export
## Summary

`design.md lint` should warn when frontmatter contains token-like fields that will be ignored by exporters.

Right now, a DESIGN.md file can contain explicit design-token data, pass lint, and then export incomplete Tailwind output because the token-like fields are not part of the recognized schema.

## Why this matters

My use case is to have an LLM generate a DESIGN.md file, then use `design.md lint` to tell the agent how to fix it before exporting tokens.

A successful lint plus successful export implies the exported Tailwind theme represents the DESIGN.md. In this case, export silently drops the color palette, so downstream agents or build tools may trust an incomplete Tailwind config.

## Reproduction

Given a DESIGN.md file like:

```yaml
---
brand_name: "Example"
base_colors:
  ink: "#0B0F14"
  paper: "#F6F7FA"
semantic_colors:
  background: "#F6F7FA"
  foreground: "#0B0F14"
typography:
  body:
    fontFamily: "Inter"
---
```

Run:

```bash
npx --yes @google/design.md@0.2.0 lint --format json DESIGN.md
npx --yes @google/design.md@0.2.0 export --format json-tailwind DESIGN.md
```

Observed export:

```json
{
  "theme": {
    "extend": {
      "colors": {},
      "fontFamily": {
        "body": ["Inter"]
      }
    }
  }
}
```

## Expected behavior

`lint` should warn that `base_colors` and `semantic_colors` are not recognized token fields and will not be exported.

More generally, lint should warn when unsupported frontmatter fields look like design tokens, such as:

- maps containing hex color values
- maps containing typography properties like `fontFamily`
- maps containing dimensions or spacing-like values
- DTCG-like token objects with `$value` / `$type`

## Version

`@google/design.md@0.2.0`

mvanhorn:
## Summary

Adds a new `token-like-ignored` lint rule that warns when a top-level frontmatter key looks like a design-token map but is not part of the recognized export schema. Currently, keys like `base_colors` or `semantic_spacing` pass lint silently and are then completely dropped by `design.md export`. This is especially confusing for LLM-generated DESIGN.md files, where lint success implies export completeness but does not guarantee it.

## Why this matters

Fixes #100. The reporter's full color palette was silently discarded because their LLM used `base_colors` instead of `colors`. A warning at lint time surfaces the problem before export swallows the data.

## Approach

The rule detects "token-like" maps by inspecting leaf values for hex color strings, CSS dimension strings, and typography property names (`fontFamily`, `fontSize`, etc.). A plain metadata object like `meta: { author: "Jane" }` stays silent; a map of hex colors or rem values triggers a warning pointing the author toward the correct schema key.

To make raw YAML values available at rule evaluation time, the parser now carries `rawValues` through to `ParsedDesignSystem`, and the model layer extracts `unknownKeyValues` for keys not in the schema. No changes to the export pipeline are needed. Ten unit tests cover the happy path, edge cases (flat scalars, non-token objects, nested maps), and multi-key scenarios.

Fixes #100

### DIFF
--- a/packages/cli/src/commands/spec.test.ts
+++ b/packages/cli/src/commands/spec.test.ts
@@ -89,6 +89,6 @@ describe('spec command', () => {
     const output = JSON.parse(outputStr);
     expect(output.spec).toBeDefined();
     expect(output.rules).toBeDefined();
-    expect(output.rules.length).toBe(9);
+    expect(output.rules.length).toBe(10);
   });
 });
--- a/packages/cli/src/linter/index.ts
+++ b/packages/cli/src/linter/index.ts
@@ -45,6 +45,7 @@ export {
   missingSections,
   missingTypography,
   unknownKey,
+  tokenLikeIgnored,
 } from './linter/rules/index.js';
 export { contrastRatio } from './model/handler.js';
 export { TailwindEmitterHandler } from './tailwind/handler.js';
--- a/packages/cli/src/linter/linter/rules/index.ts
+++ b/packages/cli/src/linter/linter/rules/index.ts
@@ -24,6 +24,7 @@ import { missingSectionsRule } from './missing-sections.js';
 import { sectionOrderRule } from './section-order.js';
 import { missingTypographyRule } from './missing-typography.js';
 import { unknownKeyRule } from './unknown-key.js';
+import { tokenLikeIgnoredRule } from './token-like-ignored.js';
 
 /** The default set of lint rule descriptors, in order. */
 export const DEFAULT_RULE_DESCRIPTORS: RuleDescriptor[] = [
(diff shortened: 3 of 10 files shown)


davideast:
Thank you @mvanhorn!

mvanhorn:
Thanks for merging this, @davideast. The token-like-ignored lint rule should catch those silently dropped frontmatter keys early.

mvanhorn:
Thanks for getting this in, @davideast. The token-like-ignored lint should save people from frontmatter keys that silently fall out of export.

PR #79, opened 2026-05-14
fix(cli): prevent crash on non-string YAML properties (Issue #75)

SyedaQurratAI:
Fixed the "raw.match is not a function" crash (Issue #75) by adding type guards for non-string YAML properties.

Changes:

Added typeof checks in parseDimensionParts and parseCssColor.

Updated ModelHandler to handle numeric/boolean values gracefully.

PR #107, opened 2026-06-15
fix(cli): add defensive type guard to parseDimension to prevent runtime crashes

### LINKED ISSUE #106: v0.2.0 published build crashes with "raw.match is not a function" — parseDimensionParts missing type guard
## Summary

The published `@google/design.md@0.2.0` build (in `dist/linter/index.js:20861`) is missing the `typeof raw !== 'string'` type guard that exists in the source on `main`. This causes the linter to crash with `"Unexpected error during model building: raw.match is not a function"` when YAML contains non-string values in `spacing`.

## Steps to reproduce

1. Create a minimal DESIGN.md with `spacing: { unit: 8 }` (YAML parses `8` as a JS number, not a string)
2. Run `npx @google/design.md lint DESIGN.md`
3. Observe: `{ "findings": [{ "severity": "error", "message": "Unexpected error during model building: raw.match is not a function" }] }`

## Root cause

**Source on `main`** (`packages/cli/src/linter/model/spec.ts:147-153`):
```ts
export function parseDimensionParts(raw: string): { value: number; unit: string } | null {
  if (typeof raw !== 'string') return null;   // ← type guard present
  const match = raw.match(/^(-?\d*\.?\d+)([a-zA-Z%]+)$/);
  ...
}
```

**Published v0.2.0 build** (`dist/linter/index.js:20861-20867`):
```js
function parseDimensionParts(raw) {
  const match = raw.match(/^(-?\d*\.?\d+)([a-zA-Z%]+)$/);  // ← NO type guard
  ...
}
```

The type guard was removed/lost in the published build. `raw.match` crashes whenever `raw` is a number, array, or other non-string value.

## Trigger conditions in DESIGN.md

Any non-string scalar in `spacing` triggers the crash. The canonical Stitch schema expects quoted dimension strings (`unit: "8px"`, not `unit: 8`), so this is reproducible with any non-conforming input — including values that the linter's own `SCHEMA_KEYS` says are valid for the `spacing` key.

## Fix

Restore the type guard at the published build level. Either:
1. Re-add the guard in the TS source (already present in `main`) and re-publish v0.2.1, **or**
2. Add a guard at every call site that passes arbitrary `raw` into `parseDimensionParts` / `isParseableDimension` / `isValidColor` (the `spacing` handler at `handler.ts:115-126` is the immediate culprit — it lacks the `typeof raw === 'string'` check that `rounded` and `colors` handlers have)

## Workaround for users

Until fixed, quote all `spacing` values as strings:
```yaml
spacing:
  unit: "8"      # not 8
  "1": "4px"     # not 1: 4px
```

## Environment

- `@google/design.md@0.2.0`
- Node 22+
- macOS / Linux

SyedaQurratAI:
Fixes #106

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2625 @@
+{
+  "name": "design-monorepo",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "design-monorepo",
+      "workspaces": [
+        "packages/*"
+      ],
+      "devDependencies": {
+        "@types/node": "^25.6.0",
+        "bun-types": "^1.3.12",
+        "turbo": "latest",
+        "typescript": "latest"
+      }
+    },
+    "node_modules/@alloc/quick-lru": {
+      "version": "5.2.0",
+      "resolved": "https://registry.npmjs.org/@alloc/quick-lru/-/quick-lru-5.2.0.tgz",
+      "integrity": "sha512-UrcABB+4bUrFABwbluTIBErXwvbsU/V7TZWfmbgJfbkwiBuziS9gxdODUyuiecfdGQ85jglMW6juS3+z5TsKLw==",
+      "dev": true,
+      "license": "MIT",
+      "engines": {
+        "node": ">=10"
+      },
+      "funding": {
+        "url": "https://github.com/sponsors/sindresorhus"
+      }
+    },
+    "node_modules/@google/design.md": {
+      "resolved": "packages/cli",
+      "link": true
+    },
+    "node_modules/@jridgewell/gen-mapping": {
+      "version": "0.3.13",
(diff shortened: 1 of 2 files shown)


SyedaQurratAI:
The local tests pass (272 passed), but conformance.test.ts fails and times out due to a 404 network resolution error on @terrazzo/token-types@^2.4.0 from npm registry. This seems to be an upstream environment issue.

davideast:
Thank you @SyedaQurratAI! I fixed the terrazo issue so this should hopefully pass.

davideast:
Hey @SyedaQurratAI, thank you for jumping on this. I just realized that this is already fixed. Your earlier PR #79 added the type guard to `parseDimensionParts`, which is the right place for it since all callers benefit from the check there. That fix was on main but hadn't been published yet when #106 was filed. We just shipped it in v0.3.0 (PR #108), so the published build now includes your fix.

Going to close this one since the issue is resolved. Please let me know if I'm mistaken though! Thank you for both contributions!

PR #117, opened 2026-06-25
feat: data-driven Color and Dimension type definitions (fixes #97)

### LINKED ISSUE #97: Color type definition is not driven by spec-config.yaml
The spec-config.yaml header says:

> This file is the single source of truth for the DESIGN.md format specification.

But the Color type definition is hardcoded as prose in `spec.mdx`, not derived from the yaml config. This means `spec-config.yaml` is the source of truth for sections, units, typography properties, component sub-tokens, and recommended token names, but not for what a `Color` actually is.

The typography properties are fully data-driven from the yaml (`typography_properties` list), and the spec generator renders them dynamically. The Color type definition should follow the same pattern.

### Current state

- `spec-config.yaml` references `type: Color` in `component_sub_tokens` but never defines what `Color` means
- `spec.mdx` has the Color definition as a static paragraph (line 46)
- If someone reads only `spec-config.yaml` to understand the spec, they would not know what color formats are accepted

### Suggestion

Add a `types` section to `spec-config.yaml` that defines the primitive types (`Color`, `Dimension`, etc.) with their descriptions and accepted formats. Then update the spec generator to render these type definitions from the config instead of hardcoding them in the mdx template.

Something like:

```yaml
types:
  Color:
    description: "A color value is any valid CSS color string."
    formats:
      - "Hex: #RGB, #RGBA, #RRGGBB, #RRGGBBAA"
      - "Named colors: red, cornflowerblue, transparent"
      - "Functional: rgb(), rgba(), hsl(), hsla(), hwb()"
      - "Wide-gamut: oklch(), oklab(), lch(), lab()"
      - "Mixing: color-mix(in srgb, ...)"
    note: "All color values are internally converted to sRGB for WCAG contrast checking."
  Dimension:
    description: "A dimension value is a string with a unit suffix."
    units: [px, em, rem]
```

### Related

- #53: The Color type definition was just updated in PR #96 to document all supported CSS color formats. That change was made directly in the mdx template because there was no other mechanism.

Emp1500:
Fixes #97.

  `spec-config.yaml` claims to be the single source of truth for the DESIGN.md spec, but the `Color` and `Dimension` type definitions were hardcoded as
  static prose in `spec.mdx`. This PR makes them fully data-driven, following the same pattern already used for `typography_properties`.

  ## Changes

1) **`spec-config.yaml`** — new `types:` section with `Color` (description, formats, note) and `Dimension` (description only; units reuse the top-level `units:` array to avoid duplication).
 2) **`spec-config.ts`** — `TypeDefSchema`, `TypeDef` interface, `PRIMITIVE_TYPES` export; `types` is a required field so a missing key is caught at load time.
 3)**`renderers.ts`** — `typeDefinitions(config, typeName?)` renderer; the optional `typeName` argument lets Color and Dimension be emitted at their original positions in the spec.
 4) **`spec.mdx`** — hardcoded Color/Dimension prose replaced with `{typeDefinitions('Color')}` `{typeDefinitions('Dimension')}`, preserving the original order: Color → typography properties → Dimension
 5)**`docs/spec.md`** — regenerated; output is identical in content to before.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -59,9 +59,7 @@ components:
 
 The `<scale-level>` placeholder represents a named level in a sizing or spacing scale. Common level names include `xs`, `sm`, `md`, `lg`, `xl`, and `full`. Any descriptive string key is valid.
 
-**Color**: A color value is any valid CSS color string.
-
-Supported formats include:
+**Color**: A color value is any valid CSS color string. Supported formats include:
 
 - Hex: `#RGB`, `#RGBA`, `#RRGGBB`, `#RRGGBBAA`
 - Named colors: `red`, `cornflowerblue`, `transparent`
@@ -73,10 +71,6 @@ All color values are internally converted to sRGB for WCAG contrast checking. Th
 
 Hex notation (`#RRGGBB`) remains the recommended default for simplicity and broad tooling support.
 
-**Dimension**: A dimension value is a string with a unit suffix.
-
-Valid units are: px, em, rem.
-
 - `fontFamily` (string)
 - `fontSize` (Dimension)
 - `fontWeight` (number) - A numeric font weight value (e.g., `400`, `700`). In YAML, this may be expressed as either a bare number or a quoted string; both are equivalent.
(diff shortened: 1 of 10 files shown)


Emp1500:
I wanted to give a bit of context on the approach here since the implementation has a few moving parts.

The core idea was to follow the exact same pattern that typography_properties already uses — data in the YAML, rendered dynamically by a function. So typeDefinitions() in renderers.ts works the same way typographyPropertyList() does.

One thing worth calling out:  Dimension doesn't have a formats list in the YAML because its units are already defined in the top-level units: array. I debated whether to duplicate them inside types. Dimension to make the type self-contained, but decided against it if they lived in both places, someone could update one and silently leave the other stale. The top-level units: array is already the authoritative list (the linter reads from it), so having Dimension reference the same source felt more honest. That said, if you'd prefer the type definition to be self-contained and are okay with a small duplication, happy to go that route instead.

I also kept the original reading order from spec.mdx — Color appears before the typography properties list, and Dimension appears after it. That meant giving typeDefinitions() an optional typeName argument so the two types could be emitted at their original positions rather than grouping them together.

rajpratham1 (approved):
This PR is a substantial refactor that centralizes primitive type definitions into the configuration layer instead of duplicating them across documentation, renderers, and generated spec files. The overall direction improves maintainability, reduces duplication, and makes future specification changes significantly easier. The accompanying tests also provide good coverage for the new rendering behavior.

davideast (approved):
LGTM! Thank you @Emp1500!

PR #120, opened 2026-06-25
fix(color): correct grad hue parsing and reject non-percentage color-mix weights

rmyndharis:
## Summary

Two CSS color-parsing correctness fixes in `model/color-parser.ts`.

### `grad` hue angles were parsed as radians

`parseHue` checked `endsWith('rad')` before `endsWith('grad')`. Since `"100grad"` ends with `"rad"`, it matched the radians branch first — `parseFloat("100g") * 180/π % 360 ≈ 329.58°` instead of `100grad = 90°`. The token was accepted with no error, so the wrong hue silently propagated to the resolved hex, WCAG luminance/contrast result, and the Tailwind/DTCG exports. Affects every hue notation: `hsl`, `hwb`, `lch`, `oklch`.

Fix: test `grad` before `rad` (longest-suffix-first). `deg`/`rad`/`turn`/unitless behavior is byte-identical.

### `color-mix` accepted and over-scaled bare-number weights

`parseColorWithWeight` treated any numeric token as a weight and multiplied bare numbers by 100. `color-mix(in srgb, red 20, blue)` became weight `2000`, producing a wildly wrong (red-dominated) blend instead of being rejected. CSS `color-mix` weights are percentages only.

Fix: require a `%` suffix; a bare number is no longer a weight, so such input returns `null` (invalid color) — matching CSS.

## Testing

- `bun test`: **284 pass, 1 skip, 0 fail** (added 2 tests: `100grad === 90deg`, and bare-number color-mix weight rejection)
- Existing `%`-weight color-mix and `hsl`/`hwb`/`lch`/`oklch` tests unchanged and passing

### DIFF
--- a/packages/cli/src/linter/model/color-parser.ts
+++ b/packages/cli/src/linter/model/color-parser.ts
@@ -306,10 +306,10 @@ function parseHue(s: string): number {
   let val = 0;
   if (lower.endsWith('deg')) {
     val = parseFloat(lower.slice(0, -3));
-  } else if (lower.endsWith('rad')) {
-    val = (parseFloat(lower.slice(0, -3)) * 180) / Math.PI;
   } else if (lower.endsWith('grad')) {
     val = (parseFloat(lower.slice(0, -4)) * 360) / 400;
+  } else if (lower.endsWith('rad')) {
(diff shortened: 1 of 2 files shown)


davideast (approved):
Thank you again @rmyndharis!

PR #121, opened 2026-06-25
perf: bound token-validation cost on adversarial input

rmyndharis:
## Summary

The CLI parses arbitrary user-supplied markdown/YAML (file or stdin). Three validation paths can be driven into super-linear CPU or unbounded recursion by a crafted DESIGN.md. Each fix is a small guard that leaves results for legitimate input unchanged.

### 1. Quadratic backtracking in dimension validation
`parseDimensionParts` (`model/spec.ts`) uses `/^(-?\d*\.?\d+)([a-zA-Z%]+)$/`; the same shape is in `token-like-ignored`'s `CSS_DIMENSION_RE`. On a long all-digit string the engine backtracks every split point (clean O(n²) — measurable seconds at ~80k chars). Real CSS dimensions are a handful of characters, so a leaf value over 64 chars is rejected before the regex runs.

### 2. Unbounded Levenshtein cost for unknown keys
`unknown-key` builds a full `(m+1)×(n+1)` DP matrix against every schema key for each unknown top-level key, with no length guard. Since edit distance is always ≥ the length difference, a key whose length differs from a schema key by more than `MAX_TYPO_DISTANCE` can never be a typo — those comparisons are now skipped. Suggestions are provably unchanged.

### 3. Unbounded `color-mix()` recursion
`parseCssColor` recurses into each inner color of a `color-mix()` with no depth bound; a deeply nested value throws `RangeError: Maximum call stack size exceeded`, which the model's catch-all turns into a single generic error that discards every other finding for the file. A depth counter (cap 32) makes an over-deep value resolve to a normal "invalid color" finding instead, leaving the rest of the model intact.

## Testing

- `bun test`: **285 pass, 1 skip, 0 fail** (added 3 tests). The suite completes in ~0.3s, including a 100k-char dimension, a 50k-char unknown key, and a 50-deep `color-mix` — all previously slow/throwing, now instant/graceful.
- `bun run lint` (`tsc --noEmit`): clean.
- The new color-mix test asserts the over-deep value is rejected per-token **and** that a sibling valid color still resolves (i.e. no model-wide collapse).

### DIFF
--- a/packages/cli/src/linter/linter/rules/token-like-ignored.ts
+++ b/packages/cli/src/linter/linter/rules/token-like-ignored.ts
@@ -26,6 +26,14 @@ const HEX_COLOR_RE = /^#([0-9a-fA-F]{3,4}|[0-9a-fA-F]{6}|[0-9a-fA-F]{8})$/;
 const CSS_DIMENSION_RE = /^-?\d*\.?\d+[a-zA-Z%]+$/;
 
 /**
+ * Upper bound on a token-like leaf value's length before pattern matching.
+ * A hex color or CSS dimension is short; longer strings cannot match either,
+ * so capping the length avoids pathological regex backtracking on oversized
+ * attacker-supplied values.
+ */
+const MAX_TOKEN_VALUE_LENGTH = 64;
+
+/**
  * Typography-flavored property names that strongly suggest this map holds
  * design tokens rather than arbitrary metadata.
  */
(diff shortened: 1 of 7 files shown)


rajpratham1 (approved):
This is a valuable hardening PR that improves the linter's resilience against adversarial or pathological inputs without changing behavior for valid DESIGN.md files. The changes are narrowly scoped, well documented, and backed by targeted regression tests covering each new safeguard.

davideast (approved):
Thank you again (and again) @rmyndharis

PR #122, opened 2026-06-25
fix(export): escape line terminators in Tailwind v4 CSS string values

rmyndharis:
## Summary

The Tailwind v4 exporter wraps font-family values with `cssStringLiteral`, which escaped only `\` and `"`. A value containing a raw newline / carriage return / form feed — valid via a YAML double-quoted or block scalar in a DESIGN.md — was emitted verbatim inside the `@theme` CSS string literal. Raw line terminators are illegal inside a CSS string and can break the value out of the quoted token, producing malformed CSS.

This complements the existing token-name validation (which already rejects non-identifier names): names are validated, and string *values* are now fully escaped.

## Fix

Escape `\n`, `\r`, and `\f` as CSS hex escapes (e.g. `\a `) in `cssStringLiteral`, alongside the existing `\`/`"` escaping.

## Testing

- `bun test`: **283 pass, 1 skip, 0 fail** (added a test: a font-family containing a newline emits `"Evil\a Family"` with no raw newline)
- Existing quote/backslash escaping test unchanged and passing

### DIFF
--- a/packages/cli/src/linter/tailwind/v4/handler.test.ts
+++ b/packages/cli/src/linter/tailwind/v4/handler.test.ts
@@ -99,6 +99,20 @@ describe('TailwindV4EmitterHandler', () => {
       if (!result.success) throw new Error('Expected success');
       expect(result.data.theme.fontFamily?.['fancy']).toBe('"Fancy \\"Font\\""');
     });
+
+    it('escapes line terminators in font-family values', () => {
+      const state = buildState({
+        typography: {
+          multi: { fontFamily: 'Evil\nFamily', fontSize: '16px' },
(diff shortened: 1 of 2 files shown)


rajpratham1 (approved):
This is a well-scoped bug fix that improves the correctness of the generated Tailwind v4 CSS. Escaping line terminators (\n, \r, and \f) as CSS hex escapes prevents invalid CSS string literals while preserving the intended value. The accompanying regression test clearly covers the reported edge case and verifies that raw newlines no longer appear in the emitted output.

davideast (approved):
Thank you @rmyndharis!

PR #124, opened 2026-06-25
chore: declare supported Node engine range (>=18)

rmyndharis:
## Summary

The published package has no `engines.node`, so consumers get no signal about the minimum supported Node version. The CLI/library uses modern Node features — ESM with `import.meta.url` + `fileURLToPath`, an `exports` map, and async iteration over `process.stdin` — so a floor is appropriate.

Declares `engines.node: ">=18.0.0"`. This is advisory by default (npm warns below it; it does not hard-block unless the consumer sets `engine-strict`), so it is non-breaking.

## Testing

Metadata only. `package.json` validated as parseable; `node`/`bun` resolution unaffected.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -19,6 +19,9 @@
   ],
   "private": false,
   "type": "module",
+  "engines": {
+    "node": ">=18.0.0"
+  },
   "bin": {
     "design.md": "./dist/index.js",
     "designmd": "./dist/index.js"


rajpratham1 (approved):
This is a small but useful maintenance change. Declaring the supported Node.js engine in package.json helps package managers warn users when they're using an unsupported runtime and makes the project's runtime requirements explicit. Since it doesn't change application logic or dependencies, it's a safe improvement.

PR #125, opened 2026-06-25
feat(lint): flag unrecognized typography sub-properties

rmyndharis:
## Summary

Typography sub-properties outside the schema — `fontFamily`, `fontSize`, `fontWeight`, `lineHeight`, `letterSpacing`, `fontFeature`, `fontVariation` — were silently dropped by the model: never resolved, never emitted to Tailwind/DTCG, and with **no finding**. A typo (`fontwight`) or an unsupported property (`textTransform`, `color`) simply vanished, looking identical to "the linter didn't notice."

This mirrors the existing treatment of unknown **component** sub-tokens, which the `broken-ref` rule already reports as a warning (*"'X' is not a recognized component sub-token"*).

## Change

`parseTypography` now emits a `warning` for each sub-property not in the schema, e.g.:

```
'textTransform' is not a recognized typography property. Valid properties: fontFamily, fontSize, fontWeight, lineHeight, letterSpacing, fontFeature, fontVariation.
```

Recognized properties resolve exactly as before. Warnings don't affect the exit code (only errors do), so this is additive.

## Testing

- `bun test`: **283 pass, 1 skip, 0 fail** (added a test: an unknown `textTransform` warns at `typography.<token>.textTransform` while `fontFamily` still resolves and never warns).
- `bun run lint` (`tsc --noEmit`): clean.
- Verified no existing fixture or test uses a non-standard typography property, so no current case newly warns.

### DIFF
--- a/packages/cli/src/linter/model/handler.test.ts
+++ b/packages/cli/src/linter/model/handler.test.ts
@@ -366,6 +366,20 @@ describe('ModelHandler', () => {
       const headline = result.designSystem.typography.get('headline');
       expect(headline?.fontWeight).toBe(700);
     });
+
+    it('warns about unrecognized typography sub-properties that are silently dropped', () => {
+      const result = handler.execute(makeParsed({
+        typography: {
+          'headline': { fontFamily: 'Inter', textTransform: 'uppercase' },
+        },
+      }));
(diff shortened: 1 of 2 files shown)


davideast (approved):
LGTM! Thank you (again) @rmyndharis!

PR #126, opened 2026-06-25
fix(export): exit 0 on a successful export regardless of source lint findings

rmyndharis:
## Summary

`export` set its exit code from the *source's* lint summary:

```js
process.exitCode = report.summary.errors > 0 ? 1 : 0;
```

So exporting a DESIGN.md that has a lint **error** (e.g. one invalid color) exited `1` even though the export ran and produced correct output for every valid token. That conflates two different things — "the source has lint findings" and "the export failed" — in one exit code, and it's undocumented.

## Change (behavior change)

Decouple the export exit code from source lint findings:

- **0** — export succeeded (even if the source has lint findings; use `lint` to gate on those)
- **1** — invalid `--format` or an emitter failure
- **2** — input file could not be read (`readInput`)

Documented in the README's `export` section. This is a deliberate behavior change on the success path (export of a lint-error file: `1` → `0`); flagging it explicitly in case the coupling was intended as a validation gate — `lint` already serves that purpose.

## Testing

- New `commands/export-exit-code.test.ts` (spawns the CLI): export of a lint-error file → exit `0` with output; `lint` of the same file → exit `1`.
- `bun test`: **284 pass, 1 skip, 0 fail**. `bun run lint`: clean.
- Verified against the built CLI with both an error-containing and a valid source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,8 @@ npx @google/design.md export --format dtcg DESIGN.md > tokens.json
 | `tailwind` | JSON | Alias for `json-tailwind` |
 | `dtcg` | JSON | W3C Design Tokens Format Module |
 
+Exit code `0` on a successful export (regardless of any lint findings in the source — run `lint` to gate on those), `1` on an invalid `--format` or an emitter error, and `2` if the input file cannot be read.
+
 ### `spec`
 
 Output the DESIGN.md format specification (useful for injecting spec context into agent prompts).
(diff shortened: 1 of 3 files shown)


rajpratham1 (approved):
This is a sensible fix that clearly separates the responsibilities of the export and lint commands. If an export successfully produces output, it should exit with 0 regardless of lint findings, while lint remains the command responsible for enforcing validation failures. The added regression tests cover both behaviors, and the README has been updated to document the exit code semantics.

rajpratham1 (approved):
This change corrects the CLI behavior by separating export success from lint validation. If an export is successfully generated, it should return a success exit code regardless of validation findings, while lint remains the command responsible for enforcing validation failures. The implementation is small, well-scoped, and backed by appropriate regression tests and documentation updates.

davideast (approved):
Thank you yet again (and again) @rmyndharis!

Issue #13, opened 2026-04-21
Support for different modes/themes?


craigsmitham:
Thanks for promoting an open standard for DESIGN.md. Our token set has ~100 role tokens maintained in parallel for light and dark. What if the spec were extended to support something like this?

  ```yaml
  ---
  version: alpha
  name: AcmeOrg
  themes: [light, dark]
  default-theme: light
  colors:
    surface:    { light: "#ffffff", dark: "#1b1a18" }
    on-surface: { light: "#2a2825", dark: "#f3f0ec" }
    brand-accent: "#5f3dc4"
  components:
    button: { backgroundColor: "{colors.surface}" }   # resolves per active theme
  ---
```

**Comment by dd-bmunge** [comment]
Strong +1 on this. We run a multi-platform design system with several brand themes, each with light/dark modes, across thousands of tokens organized in three tiers (base primitives, semantic usage tokens, component tokens).

The inline `{ light: ..., dark: ... }` approach in the OP works well for light/dark within a single theme. For brand-level theming on top of that, I think there also needs to be a multi-file inheritance model. Something like:

```yaml
---
name: Consumer
extends: ./DESIGN.base.md
modes:
  light:
    colors:
      text-default: "{colors.neutral-100}"
      background-default: "{colors.white}"
  dark:
    colors:
      text-default: "{colors.neutral-0}"
      background-default: "{colors.neutral-900}"
---
```

Where `DESIGN.base.md` defines the primitive tokens and defaults, and brand files extend it with overrides. This would give you:

1. A base file defining primitives and defaults
2. Theme files that extend the base and override specific tokens
3. Mode definitions (light/dark) within each theme file
4. `export` and `lint` resolving the full inheritance chain

Without something like this, design.md can't be used as a source of truth for any design system with brand variants -- which is most production systems at scale.

MickaelV0:
Adding a real-world case that stretches the theme model beyond light/dark — a **branded aesthetic variant** that shares brand hues but differs in ways a theme token can't express.

Our product "Lyra" ships **two aesthetic variants**:

| Axis | v1 Obsidian | v2 GitHub-dark |
|---|---|---|
| Scope | Marketing / hero / avatar / READMEs | Dep-graphs / architecture / dense docs |
| `bg` | `#0a0a0f` | `#0d1117` |
| Elevation ladder | `bg → surface` (2 levels, recessed) | `bg → cell → panel → card` (4 levels, raised) |
[...]

PR #128, opened 2026-06-26
feat: multi-theme support via per-token mode values (light/dark)

mvanhorn:
## Summary
Adds multi-theme support to the DESIGN.md spec via inline per-token mode values: a single color/dimension token can carry `{ light, dark }` (and other declared modes) instead of being duplicated across parallel light/dark tokens. This is the maintainer-endorsed `modes` shape from #13.

## Why this matters
The reporter in #13 maintains ~100 role tokens duplicated in parallel for light and dark. @davideast is "very supportive of multi-themes," explicitly endorsed the inline `modes` approach because "it will be agent friendly when tooling isn't present," and listed it "at the top of the list." This implements that endorsed shape only, additively and backward-compatibly: existing scalar-only DESIGN.md files parse, resolve, and export exactly as before.

## Changes
- Parser (`parser/spec.ts`, `parser/handler.ts`): accept a mode-keyed object value (`{ light, dark }`) alongside the existing scalar form; capture `themes` / `default-theme` frontmatter keys (registered in `spec-config.ts`).
- Model (`model/spec.ts`, `model/handler.ts`): resolve each token per-mode while keeping single-value resolution as the default-theme path.
- Tailwind v4 emitter (`tailwind/v4/serialize.ts`): emit the default mode in `:root` and alternate modes under a `prefers-color-scheme` / data-attribute selector.

## Scope
Covers the inline per-token `modes` model (parse + resolve + export) only. The `extends` / multi-file inheritance and external-file token sourcing the maintainer floated separately are intentionally out of scope and need their own spec decision.

## Testing
`bun test src/linter` -> 274 pass, 1 skip, 0 fail (621 assertions across 30 files). New tests cover `{ light, dark }` parse/resolve, default-theme backward compatibility, Tailwind v4 export of default + alternate modes, and malformed-mode-object findings.

Refs #13

AI was used for assistance.

### DIFF
--- a/packages/cli/src/linter/model/handler.test.ts
+++ b/packages/cli/src/linter/model/handler.test.ts
@@ -220,6 +220,131 @@ describe('ModelHandler', () => {
     });
   });
 
+  describe('theme mode values', () => {
+    it('resolves color mode objects and keeps default-theme in the existing color map', () => {
+      const result = handler.execute(makeParsed({
+        themes: ['light', 'dark'],
+        defaultTheme: 'light',
+        colors: {
+          surface: {
+            light: '#ffffff',
+            dark: '#111111',
+          },
+        },
+      }));
+
+      expect(result.findings.filter(f => f.severity === 'error')).toHaveLength(0);
+      expect(result.designSystem.colors.get('surface')?.hex).toBe('#ffffff');
+      expect(result.designSystem.colors.has('surface.light')).toBe(false);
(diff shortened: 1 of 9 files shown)


rajpratham1 (approved):
This is a substantial feature addition that introduces first-class multi-theme (light/dark) support while preserving backward compatibility for existing single-theme design files. The implementation is comprehensive, covering parsing, model resolution, Tailwind serialization, validation, and regression tests.

PR #130, opened 2026-06-26
Update check-package.ts

krataratha:


rajpratham1 (approved):
This is a useful enhancement to the package verification script. Adding an automated production dependency audit strengthens the release validation process without affecting the package itself, and the implementation gracefully handles both JSON parsing and command failures.

PR #131, opened 2026-06-26
fix: render primitive types from spec config

### LINKED ISSUE #97: Color type definition is not driven by spec-config.yaml

juliosuas:
Fixes #97. Moves primitive type definitions into spec-config.yaml and renders them through the spec generator, so Color and Dimension are driven by the same config source as sections, units, typography properties, and component sub-tokens. Validation: focused spec-config/spec-gen tests passed; generate --check passed; full bun run test, bun run lint, bun run build, and git diff --check passed.

### DIFF
--- a/docs/spec.md
+++ b/docs/spec.md
@@ -59,17 +59,23 @@ components:
 
 The `<scale-level>` placeholder represents a named level in a sizing or spacing scale. Common level names include `xs`, `sm`, `md`, `lg`, `xl`, and `full`. Any descriptive string key is valid.
 
-**Color**: A color value is any valid CSS color string. Supported formats include:
+**Color**: A color value is any valid CSS color string.
 
-* Hex: `#RGB`, `#RGBA`, `#RRGGBB`, `#RRGGBBAA`
-* Named colors: `red`, `cornflowerblue`, `transparent`
-* Functional: `rgb()`, `rgba()`, `hsl()`, `hsla()`, `hwb()`
-* Wide-gamut: `oklch()`, `oklab()`, `lch()`, `lab()`
-* Mixing: `color-mix(in srgb, ...)`
+Supported formats include:
+
+- Hex: `#RGB`, `#RGBA`, `#RRGGBB`, `#RRGGBBAA`
+- Named colors: `red`, `cornflowerblue`, `transparent`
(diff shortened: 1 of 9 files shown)


rajpratham1 (approved):
This is a well-structured refactor that moves primitive type definitions into the shared spec configuration instead of duplicating them inside the generated documentation. It improves maintainability, keeps the specification as a single source of truth, and includes appropriate validation and regression tests.

davideast (approved):
Thank you @juliosuas!

Issue #116, opened 2026-06-25
Accessibility considerations: DESIGN.md as infrastructure for AI-generated UI


securecat:
## Summary

DESIGN.md has the potential to become foundational infrastructure for how AI coding agents produce UI. That makes it significantly different from a typical design tooling project — and it places a corresponding responsibility on the format specification itself to model accessible design practices.

This issue is not a criticism of what has been built. It is a request, rooted in optimism about this project's reach, to treat accessibility as a first-class citizen in the format spec — not an afterthought or a downstream implementation concern.

---

## The core concern: accessibility cannot be delegated to the agent

The current design of DESIGN.md is premised on the idea that a well-described visual identity, handed to an AI agent, will produce correct UI. For visual styling alone this may hold. But accessibility is not a visual-layer concern exclusively — it involves semantics, interaction patterns, state communication, and reading order. The format currently provides no guidance in these areas.

More specifically: a developer who authors a compliant DESIGN.md file today can reasonably feel they have "done their part." This creates a false sense of assurance. The agent receives a color palette and typography scale and is expected to produce accessible UI — but nothing in the format tells the agent *what accessible UI looks like*, nor does the linter verify it.

Given Google's position in the ecosystem and the scale at which this format is likely to be adopted, design decisions made here will propagate widely. This is precisely why the format should lead on accessibility — not follow.

---

### 1. Contrast checking covers only explicitly-defined component pairs

The `contrast-ratio` lint rule only fires when both `backgroundColor` and `textColor` are defined on a `components` entry. Most real-world DESIGN.md files will define color tokens without fully specifying every component — leaving the majority of color combinations unverified. A file can pass the linter with zero contrast warnings while containing combinations that fail WCAG at implementation time.

**Request:** Expand contrast validation beyond component pairs, or at minimum surface a clear warning when color tokens exist but no component-level contrast has been verified.

### 2. WCAG AA should be a baseline, not a target

The linter checks against WCAG AA (4.5:1), which is the right minimum — but the format should make clear that AA is a floor, not a destination. There is currently no path toward AAA (7:1) for body text, nor any mechanism for non-text contrast (WCAG SC 1.4.11, minimum 3:1 for UI components and graphical objects). Borders, icons, form control boundaries, and focus indicators are entirely outside the linter's scope. Treating AA as sufficient risks normalizing the minimum as the goal, particularly when the format is consumed at scale by AI agents that cannot exercise judgment about context or user needs.

[...]

PR #141, opened 2026-06-27
feat(linter): typography accessibility warnings (GH-116)

drmikecrypto:
Add lint rules for letterSpacing, lineHeight, and fontSize thresholds called out in the accessibility spec discussion.

Addresses #116 (items 7-9)


Ka77ius:
i was here

PR #143, opened 2026-06-27
feat(linter): focus-ring tokens and on-* color pairing

drmikecrypto:
## Summary
- Add `focusRingColor`, `focusRingWidth`, and `focusRingOffset` component property tokens with spec guidance (preserve browser default when omitted; WCAG 2.4.11)
- Add `focus-ring-contrast` lint rule (3:1 against backgroundColor; 2px width floor)
- Add `on-color-contrast` and `missing-on-color` rules for background/on-* token pairs
- Extend recommended color token names with `on-primary`, `on-secondary`, `on-error`

## Test plan
- [x] bun test focus-ring-contrast, missing-on-color, on-color-contrast, types, spec-config

Addresses #116 (items 4 and 5)


securecat:
Thank you for picking this up so quickly — items 4 and 5 are addressed cleanly here, and the "preserve browser default when tokens are absent" framing in item 5 is exactly the guardrail I was hoping for.

One housekeeping note: the commit message currently contains `Fixes #116`, which will cause GitHub to automatically close issue #116 when this PR is merged. However, items 1–3 and 6–10 from that issue are not addressed in this PR. It may be worth changing the reference to `Addresses #116 (items 4 and 5)` to keep the issue open for the remaining work.

Thanks again — looking forward to seeing this land.

drmikecrypto:
Good catch — updated the commit message and PR description to `Addresses #116 (items 4 and 5)` so the issue stays open for the remaining items. Applied the same `Addresses` wording on #141 and #142 for their respective item ranges as well.

PR #146, opened 2026-06-28
fix(stdin): show TTY hint when reading from interactive terminal

Emp1500:
## Problem

When a user runs a `design.md` command with `-` as the file path from an
interactive terminal, the process hangs silently with no output:

```sh
$ design.md lint -
█  ← cursor blocks here indefinitely
```

There is no indication that the tool is waiting for input, or that the user must press `Ctrl+D` to signal end-of-file. A user unfamiliar with this Unix convention has no feedback to distinguish a stalled process from one that is working correctly.

## Root Cause

`readInput` reads stdin via `for await (const chunk of process.stdin)` with no TTY detection. When stdin is an interactive terminal
(`process.stdin.isTTY === true`), the loop blocks until EOF — but nothing prompts the user to send one.

## Solution

Before entering the read loop, check whether stdin is a TTY and, if so, emit a one-line hint to `stderr`:

Piped usage is completely unaffected — `isTTY` is `false` for pipes, so the hint never appears and existing behavior is unchanged:

```sh
cat DESIGN.md | design.md lint -   # works exactly as before
```

## Design Decisions

- `readInput` now accepts an optional `stdin` stream parameter that defaults to `process.stdin`. All existing call sites remain  unchanged.
- The optional parameter makes the TTY code path fully unit-testable via injected mock streams, without any need to mock       globals.
- `StdinStream` is exported so test files can type mock streams without duplicating the interface definition.

## Changes

| File | Change |
|---|---|
| `utils.ts` | Export `StdinStream` type; add optional `stdin` param to `readInput`; add TTY check before read loop |
| `utils.test.ts` | 4 new tests: piped content, empty stream, TTY hint emission, non-TTY silence — all using injected mock streams |

### DIFF
--- a/packages/cli/src/utils.test.ts
+++ b/packages/cli/src/utils.test.ts
@@ -12,8 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { describe, it, expect } from 'bun:test';
+import { describe, it, expect, spyOn } from 'bun:test';
 import { readInput, FileReadError, formatOutput } from './utils.js';
+import type { StdinStream } from './utils.js';
+
+function makeStdin(content: string, isTTY: boolean): StdinStream {
+  async function* gen() { yield Buffer.from(content); }
+  return Object.assign(gen(), { isTTY });
+}
 
 describe('readInput', () => {
   it('throws FileReadError when file does not exist', async () => {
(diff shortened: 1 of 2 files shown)


davideast (approved):
Thank you @Emp1500! yet again that is!

PR #103, opened 2026-06-05
fix: support nested token declarations in frontmatter

vikks:
This PR fixes a bug where nested color declarations in the DESIGN.md frontmatter (e.g., background: { light: '#fff', dark: '#000' }) caused validation errors.

### Key Changes
- **Recursive Parsing**: Implemented a `forEachLeaf` helper in the `ModelHandler` to handle arbitrary token nesting depths.
- **Flattened Symbol Table**: Nested tokens are now correctly flattened (e.g., `colors.background.light`) for resolution and validation.
- **Improved Type Safety**: Updated `ParsedDesignSystem` to support nested objects in tokens.
- **Comprehensive Tests**: Added unit tests for 2, 3, and 4-level nesting, as well as mixed nested/flat declarations.

Fixes #102\n\n### Update (June 5, 2026)\n- Fixed TypeScript error by expanding `ResolvedValue` to include `number` and `boolean`.\n- Added a token nesting depth limit of 20 levels (`MAX_TOKEN_NESTING_DEPTH`) to ensure performance and prevent excessive recursion.
- Made these limits configurable via `spec-config.yaml` (`limits` section).

PR #68, opened 2026-05-03
fix(linter): support grouped color tokens

Arshgill01:
## Summary

Adds support for nested color groups under `colors` while preserving the existing flat color-token behavior.

This allows DESIGN.md files to define color scales like:

```yaml
colors:
  utility-info:
    50: "#EEF7FC"
    100: "#D4EBF7"
```

instead of requiring all tokens to be manually flattened:

```yaml
colors:
  utility-info-50: "#EEF7FC"
  utility-info-100: "#D4EBF7"
```

## What changed

* Added a recursive `ParsedColorValue` type for parsed color values.
* Updated parser/model typing to allow nested color groups without falling back to `any`.
* Added grouped color flattening in the model layer.
* Preserved flat token keys such as `utility-info-50`.
* Added symbol-table support for both:

  * `{colors.utility-info-50}`
  * `{colors.utility-info.50}`
* Updated the DESIGN.md spec documentation for grouped color tokens.
* Added tests for grouped color parsing, invalid nested color diagnostics, component references, Tailwind export, and DTCG export.

## Behavior

Grouped color input:

```yaml
colors:
  utility-info:
    50: "#EEF7FC"
```

is normalized internally/exported as:

```txt
utility-info-50
```

This keeps Tailwind-friendly output while still allowing object-path token references.

## Validation

* `bun install`
* `bun test`
* `bun run build` from `packages/cli`

Result: all 239 tests pass and the package builds successfully.

## Notes

DTCG and Tailwind exports intentionally preserve flattened token names such as `utility-info-50`. This avoids data loss and maintains compatibility with the existing flat color-token model.

Fixes #66.

PR #150, opened 2026-06-28
feat(linter): detect token name collisions for flat vs grouped keys

### LINKED ISSUE #149: feat(linter): token name collision detection for flat vs grouped keys
Following up on @davideast's suggestion in #68:

> One thing your PR has that #103 does not: collision detection for cases where a flat key and a grouped key produce the same name. That's a good idea and worth tracking. If you're interested, filing a separate issue for collision detection would be welcome and we can work on it from there. Create an issue and tag me in it.

## Problem

Since #103 merged `forEachLeaf` and arbitrary-depth nesting for `colors`, `rounded`, and `spacing`, it is now possible to author a `DESIGN.md` that produces silent name collisions in the symbol table. Two distinct cases:

**Case 1 — Duplicate dot path** (a flat key whose literal text mirrors an existing nested token's dot-path):

```yaml
colors:
  utility-info:
    50: "#EEF7FC"
  utility-info.50: "#FFFFFF"   # same symbol table key: colors.utility-info.50
```

**Case 2 — Flat/grouped name collision** (a nested token that normalizes, dots → hyphens, to the same string as an already-registered flat key):

```yaml
colors:
  utility-info-50: "#EEF7FC"   # registers as utility-info-50
  utility-info:
    50: "#FFFFFF"               # also normalizes to utility-info-50 in flat output
```

In both cases the second value silently overwrites the first in the symbol table. The Tailwind v4 emitter then exports whichever value happened to win the race — with no diagnostic emitted to the user.

The same issue applies to `rounded` and `spacing` since they also use `forEachLeaf` as of #103.

## Proposed solution

During `ModelHandler` Phase 1 (primitive token resolution), track two sets per token category (`colors`, `rounded`, `spacing`):

- `seenKeys: Set<string>` — exact dot-separated path, to detect Case 1
- `seenNormalized: Map<string, string>` — hyphen-normalized name, to detect Case 2

Emit an `error`-severity finding for each violation *before* writing to the symbol table, and skip the offending token so the winning value is deterministic and the user sees the diagnostic at lint time.

@davideast — tagging you as requested in #68.

Arshgill01:
Closes #149

Requested by @davideast in #68:
> One thing your PR has that #103 does not: collision detection for cases where a flat key and a grouped key produce the same name. That's a good idea and worth tracking.

## What this does

`forEachLeaf` (merged in #103) enables arbitrary nesting depth for `colors`, `rounded`, and `spacing` tokens. As a side-effect it is now possible to write a `DESIGN.md` where two differently-structured keys resolve to the same symbol table entry with no diagnostic emitted. Two collision shapes exist:

### Case 1 — Duplicate dot path
A flat string key whose literal value matches an existing nested token's dot-separated path:

```yaml
colors:
  utility-info:
    50: "#EEF7FC"
  utility-info.50: "#FFFFFF"   # → error: Duplicate token path 'colors.utility-info.50' detected.
```

### Case 2 — Flat/grouped name collision
A nested token whose path, when dots are replaced by hyphens, matches a flat key that is already registered:

```yaml
colors:
  utility-info-50: "#EEF7FC"
  utility-info:
    50: "#FFFFFF"              # → error: Grouped color token flattens to 'utility-info-50', which is already defined.
```

In both cases the offending token is **skipped** (not written to the symbol table) and an `error`-severity finding is emitted so the user sees it at lint time, before the Tailwind v4 emitter processes the output. Behaviour is consistent across all three token categories.

## Changes

| File | Change |
|---|---|
| `packages/cli/src/linter/model/handler.ts` | Adds `seenKeys` (`Set<string>`) and `seenNormalized` (`Map<string, string>`) tracking inside the `colors`, `rounded`, and `spacing` Phase 1 `forEachLeaf` callbacks |
| `packages/cli/src/linter/model/handler.test.ts` | 6 new test cases covering both collision types for all three token categories |

## Verification

```
bun test      → 288 pass, 0 fail
bun run lint  → tsc clean
bun run build → Bundled successfully
```

### DIFF
--- a/packages/cli/src/linter/model/handler.ts
+++ b/packages/cli/src/linter/model/handler.ts
@@ -54,7 +54,10 @@ export class ModelHandler implements ModelSpec {
       // ── Phase 1: Resolve primitive tokens ──────────────────────────
       // Colors
       if (input.colors) {
+        const isCollision = buildCollisionGuard('colors', findings);
         forEachLeaf(input.colors, (name, raw) => {
+          if (isCollision(name)) return;
+
           if (typeof raw === 'string' && isTokenReference(raw)) {
             // Store raw reference for later resolution
             symbolTable.set(`colors.${name}`, raw);
@@ -85,7 +88,10 @@ export class ModelHandler implements ModelSpec {
 
       // Rounded
       if (input.rounded) {
+        const isCollision = buildCollisionGuard('rounded', findings);
         forEachLeaf(input.rounded, (name, raw) => {
+          if (isCollision(name)) return;
+
           if (typeof raw === 'string') {
             if (isParseableDimension(raw)) {
               const resolved = parseDimension(raw);
@@ -114,7 +120,10 @@ export class ModelHandler implements ModelSpec {
 
       // Spacing
       if (input.spacing) {
+        const isCollision = buildCollisionGuard('spacing', findings);
         forEachLeaf(input.spacing, (name, raw) => {
+          if (isCollision(name)) return;
+
           if (isParseableDimension(raw)) {
             const resolved = parseDimension(raw);
             spacing.set(name, resolved);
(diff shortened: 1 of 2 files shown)


packages/cli/src/linter/model/handler.ts
Emp1500, line 1: Phase 2 can undo the collision guard. This one's a bit sneaky because it's not in your diff, but the new guard in Phase 1 interacts with the existing Phase 2 reference-resolution loop in a way that can silently corrupt the symbol table.
Phase 2 does a bare forEachLeaf walk over input.colors again with no awareness of seenKeys or seenNormalized. So in a case like this:

colors:
  utility-info-50: "{colors.brand.primary}"  # token reference
  utility-info:50: "#FFFFFF"

Phase 1 does the right thing detects the collision, emits the error, skips utility-info.50. But then Phase 2 comes along, re-walks input.colors, and when it resolves the reference for utility-info-50 it calls symbolTable.set() again  no guard, no check. Whatever Phase 1 decided gets quietly overwritten. Same issue exists in the rounded and spacing Phase 2 loops.
The straightforward fix is to not re-walk input.colors in Phase 2 at all — just iterate over symbolTable directly and resolve any entries that are still raw reference strings. That way Phase 2 only touches what Phase 1 actually registered.

Emp1500 (changes requested):
Few changes and this is ready to go:
1)Extract the collision guard into a shared helper (DRY)
2)Phase 2 re-traverses without collision guards (data corruption risk)

packages/cli/src/linter/model/handler.ts
```diff
@@ -54,7 +54,29 @@ export class ModelHandler implements ModelSpec {
...
+          if (seenNormalized.has(normalized)) {
+            findings.push({
+              severity: 'error',
+              path: `colors.${name}`,
+              message: `Grouped color token flattens to '${normalized}', which is already defined.`,
+            });
+            return;
+          }
+          seenKeys.add(name);
+          seenNormalized.set(normalized, name);
```
Emp1500: Extract the collision guard into a shared helper (DRY). The seenKeys / seenNormalized block is identical across colors, rounded, and spacing only the category string differs. Might be worth extracting before this merges, purely because typography is one refactor away from needing a fourth copy, and that's when these things tend to drift. Something like a small buildCollisionGuard(category, findings) factory that returns a (name) => boolean predicate would collapse all three blocks down to a one-liner each, with no changes to the existing messages or tests.

Emp1500 (changes requested):

Arshgill01:
Both points addressed — just force-pushed.

**1. DRY (`buildCollisionGuard` factory)**
Extracted the `seenKeys`/`seenNormalized` block into a module-level `buildCollisionGuard(category, findings)` factory that returns a `(name: string) => boolean` predicate. Each call site in `colors`, `rounded`, and `spacing` is now a single `const isCollision = buildCollisionGuard(...)` + `if (isCollision(name)) return;` — no duplicated logic, and adding `typography` later is a one-liner.

**2. Phase 2 symbol-table iteration (data-corruption fix)**
Phase 2 now iterates `symbolTable` entries directly instead of re-walking `input.*` via `forEachLeaf`. This means Phase 2 only ever touches keys that Phase 1 actually registered, so tokens that were skipped due to a collision can never be silently re-inserted. The logic is also simpler — one loop instead of three.

Emp1500 (commented):
Does the colors test still pass after the refactor?
Noticed the old colors block said "Grouped color token flattens to..." (singular), but rounded/spacing just used the category name as-is. Now everything goes through buildCollisionGuard with the same ${category} interpolation, so wouldn't calling it with 'colors' give you "Grouped colors token..." instead?

Arshgill01:
Yes, caught that during the refactor and updated the test to match the new message.

The old code had three separate inline blocks so the colors one happened to use the singular noun "color" while rounded/spacing used their category names as-is. `buildCollisionGuard` made the pattern uniform — it uses the `category` argument verbatim in both the `path` and the message string, so calling it with `'colors'` now gives:

```
Grouped colors token flattens to 'utility-info-50', which is already defined.
```

The test on line 161 was updated from `"Grouped color token..."` to `"Grouped colors token..."` to match. All 288 tests pass on the current push (`f9dbecc`).

Emp1500 (approved):
LGTM — both issues I flagged (Phase 2 corruption risk and the DRY refactor) are addressed, and the colors test fix matches the new uniform messaging. 288/288 tests passing. Nice work tightening this up.

davideast (approved):
LGTM! Thank you so much @Arshgill01!

PR #109, opened 2026-06-20
feat: add CSS custom properties export format (--format css-vars)

mvanhorn:
## Summary

Adds a `css-vars` export target so `design.md export --format css-vars DESIGN.md` emits a `:root { ... }` block of CSS custom properties derived from the token set. This makes DESIGN.md tokens directly consumable in native-CSS codebases without Tailwind or any token tooling.

## Why this matters

Issue #15 asks for a CSS variables export so DESIGN.md is useful outside the Tailwind ecosystem and gives coding agents a direct artifact to apply in native CSS. It complements the existing `dtcg` and Tailwind exports. Scope here is the core format plus `--prefix`; dark-theme / media-query variants are deferred as the issue notes.

## Changes

- New emitter under `packages/cli/src/linter/css-vars/` (`spec.ts`, `handler.ts`, `serialize.ts`) following the `DtcgEmitterHandler` pattern: a pure `execute(state)` maps `colors`, `spacing`, and `rounded` to `--<group>-<name>` declarations, and `serializeCssVars` renders the `:root` block.
- Wired into `export.ts`: `css-vars` added to the `FORMATS` enum, a new branch mirroring the other formats, and an optional `--prefix` arg threaded into the serializer (default emits `--color-...`).
- Re-exported the handler, serializer, and result type from `linter/index.ts`.
- Nested token keys flatten to dotted names (e.g. `background.light`); those dots are collapsed to hyphens so the emitted property name is valid CSS.

Sample output (`--prefix ds`):

```css
:root {
  --ds-color-surface: #f9f9ff;
  --ds-spacing-unit: 8px;
  --ds-rounded-sm: 0.25rem;
}
```

## Testing

- `bun test` in `packages/cli`: 289 pass, 1 skip, 0 fail
- `bun run lint` (tsc --noEmit): clean
- New unit tests cover colors, spacing/rounded units, `--prefix`, nested-name hyphenation, and empty input; a command-level test covers a `--format css-vars` round-trip plus the invalid-format exit-code-1 path.

Fixes #15

PR #152, opened 2026-07-06
fix(export): set exit code 0 on a successful export

arpitjain099:
The export success branches (css-tailwind, json-tailwind, dtcg, css-vars) write their output but never assign `process.exitCode`, so a successful run just leaves whatever value was already there rather than an explicit 0. This is the success-path counterpart to #126: that PR decoupled the exit code from source lint findings, but its subprocess test only covers json-tailwind, and a real process exits 0 when `exitCode` is unset, so the gap slipped through. I set `process.exitCode = 0` after the format branches (every error branch already returns first, so it only runs on success).

While confirming that, the in-process `export.test.ts` added in #109 turned out to be failing on main: it asserts `process.exitCode === 0` after a css-vars export, which was undefined. The same file also read `error.error` for the human-readable message, but the error envelope is `{ error: CODE, message: TEXT }`, so I pointed that one assertion at `error.message`.

Verified with `bun test` in `packages/cli`: the two `export.test.ts` failures now pass and the full suite is green (308 pass, 1 skip, 0 fail), `bun run lint` clean. Happy to adjust if you'd rather split the test-assertion tweak out. Thanks!

### DIFF
--- a/packages/cli/src/commands/export.test.ts
+++ b/packages/cli/src/commands/export.test.ts
@@ -66,7 +66,7 @@ describe('export command', () => {
     expect(logSpy.mock.calls.length).toBe(0);
     expect(errorSpy.mock.calls.length).toBe(1);
     const error = JSON.parse(errorSpy.mock.calls[0][0]);
-    expect(error.error).toContain('Invalid format "not-a-format"');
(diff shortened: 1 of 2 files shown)


davideast (commented):
LGTM! Thank you so much @arpitjain099!

PR #157, opened 2026-07-17
Fix/spec command path

SyedaQurratAI:
### Description
This PR resolves the `spec.md` path resolution issue in the compiled CLI. Previously, the CLI had trouble locating the specification file correctly after being built/compiled, causing command failures.

### Changes
* Fixed the path resolution logic in `spec-helpers.ts` to ensure `spec.md` is correctly located in both development and production (compiled) environments.

### Verification
* Verified that the CLI successfully locates and processes the spec file without throwing path resolution errors.
* Commit has been signed and verified using SSH keys.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,2625 @@
+{
+  "name": "design-monorepo",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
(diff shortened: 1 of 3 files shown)


davideast (commented):
Thank you @SyedaQurratAI!

PR #20, opened 2026-04-22
feat: foundational infrastructure for CSS Custom Properties export

cryptodiscolab:
# Technical Proposal: CSS Custom Properties Exporter for design.md

This proposal introduces a robust, dependency-free utility to bridge the gap between `design.md` specifications and frontend implementation by exporting design tokens directly into **CSS Custom Properties**.

## 💡 Rationale

While `design.md` successfully centralizes visual identity in a machine-readable format, frontend engineers still face manual labor when mapping these tokens to CSS. To fulfill the promise of "Design-as-Code," we require a formal automation layer that transforms YAML-defined tokens into standard web variables (`--custom-properties`).

This implementation addresses **Issue #15** by providing a portable Node.js exporter.

### Key Features
- **Zero Dependencies**: Built strictly using Node.js internal modules (`fs`, `path`) to ensure compatibility across all environments.
- **Recursive Mapping**: Seamlessly handles nested YAML structures.
- **Standardized Serialization**: Automates the creation of a `:root` scope with valid CSS syntax.
- **Comment-Aware Parsing**: Handles inline YAML comments without corrupting token values.

### Usage Pattern
```bash
node scripts/design-to-css.js <input-file> <output-file>
```

---
*Authored by the Senior Engineering Leadership at Crypto Disco DailyApp.*

PR #173, opened 2026-08-25
fix(cli): package spec for CLI command

### LINKED ISSUE #153: `spec` command fails on v0.3.0: reads dist/spec.md but the build emits dist/linter/spec.md
### Summary

The `spec` command throws `Failed to load spec.md` on a clean install of `@google/design.md@0.3.0`. The file it needs **is** in the published tarball — just at a different path than the command reads. `lint`, `diff`, and `export` are unaffected.

### Version

`@google/design.md@0.3.0` (installed from npm via `npx`/`bunx` and via `npm install`).

### Repro

```sh
npx @google/design.md@0.3.0 spec
```

### Actual

```
 ERROR  Failed to load spec.md.
  Bundled path: .../node_modules/@google/design.md/dist/spec.md
  Dev path:     .../@google/docs/spec.md
If running from a built bundle, ensure the build script copies docs/spec.md into dist/.
    at getSpecContent (.../dist/index.js:24408:11)
    at Object.run (.../dist/index.js:24474:18)
```

### Root cause

`getSpecContent()` looks for `dist/spec.md`, but the published package ships the file at **`dist/linter/spec.md`**:

```sh
$ ls node_modules/@google/design.md/dist/spec.md          # -> No such file
$ ls node_modules/@google/design.md/dist/linter/spec.md   # -> exists
```

So the build does emit `spec.md`, just under `dist/linter/` rather than `dist/`, and the command's expected path wasn't updated to match (or vice-versa).

### Suggested fix

Either point `getSpecContent()` at `dist/linter/spec.md`, or have the build copy `docs/spec.md` to `dist/spec.md` (as the error hint suggests). A packaging test that runs `design.md spec` against the built tarball would catch this class of path drift.

### Impact

Low — only the `spec` command (which prints the format reference) is broken. But it's the command agents/humans reach for to learn the format, so the failure is user-facing at an inconvenient moment.

juliosuas:
Fixes #153.

## What changed

Copy `docs/spec.md` to both bundle entry-point directories during the CLI build. The linter entry point continues to resolve `dist/linter/spec.md`, while the CLI entry point can now resolve `dist/spec.md` as expected by `getSpecContent()`.

## Verification

- Reproduced the failure on `main`: package check #20 reported `CLI spec command failed to load spec.md`.
- `bun run check-package` — 25 passed, 0 failed.
- `bun test` — 282 passed, 1 skipped, 0 failed.
- `bun run lint` — passed.
- `git diff --check` — passed.

This PR does not change `.github/workflows`. Org-level zizmor can still fail on pre-existing unpinned actions in `test.yml`; that is outside the scope of #153.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -38,7 +38,7 @@
     "access": "public"
   },
   "scripts": {
-    "build": "bun build src/index.ts src/linter/index.ts --outdir dist --target node && npx tsc --project tsconfig.build.json --emitDeclarationOnly --skipLibCheck && cp src/linter/spec-config.yaml dist/linter/ && cp src/linter/spec-config.yaml dist/ && cp ../../docs/spec.md dist/linter/",
+    "build": "bun build src/index.ts src/linter/index.ts --outdir dist --target node && npx tsc --project tsconfig.build.json --emitDeclarationOnly --skipLibCheck && cp src/linter/spec-config.yaml dist/linter/ && cp src/linter/spec-config.yaml dist/ && cp ../../docs/spec.md dist/linter/ && cp ../../docs/spec.md dist/",
     "dev": "bun run src/index.ts",
     "test": "bun test",
     "lint": "tsc --noEmit --skipLibCheck",


juliosuas:
Keeping this PR in scope of the #153 spec packaging fix (`packages/cli/package.json` only). I reverted the workflow/CI commits that hash-pinned `actions/checkout`, `oven-sh/setup-bun`, and `actions/setup-node`, added `permissions: {}`, and set `persist-credentials: false`.

Org-level zizmor (`zizmor-output`) fails on pre-existing unpinned actions in the repo's `test.yml`; that finding is not introduced by this diff. I am not expanding the PR to pin or rewrite CI.

juliosuas:
Accidentally closed when the source fork was deleted during a profile cleanup. Same commits restored here: https://github.com/google-labs-code/design.md/pull/175

PR #177, opened 2026-09-02
fix(dtcg): emit required typography lineHeight and letterSpacing

### LINKED ISSUE #172: dtcg export omits required typography properties, so output fails the schema it declares
### What happened

`export --format dtcg` emits typography tokens that are missing two of the five properties the DTCG typography value schema marks as required, so the output does not validate against the schema it declares in its own `$schema` field.

### Reproduction

Any spec with a typography section — here `olmo-earth/DESIGN.md` from [allenai/design.md](https://github.com/allenai/design.md):

```bash
npx @google/design.md@0.4.0 export --format dtcg olmo-earth/DESIGN.md | jq '.typography["body-md"]'
```

```json
{
  "$type": "typography",
  "$value": {
    "fontFamily": "Manrope, Arial, sans-serif",
    "fontSize": { "value": 16, "unit": "px" },
    "fontWeight": 400
  }
}
```

The source frontmatter for that role is:

```yaml
body-md:
  fontFamily: "{typography.font-body}"
  fontSize: 16px
  fontWeight: 400
  lineHeight: 1.6
```

### Why it's a problem

The emitted document declares `"$schema": "https://www.designtokens.org/schemas/2025.10/format.json"`. That schema's typography value definition:

```
properties: fontFamily, fontSize, fontWeight, letterSpacing, lineHeight
required:   fontFamily, fontSize, fontWeight, letterSpacing, lineHeight
additionalProperties: false
```

All five are required. The export emits three, so **every** typography token fails validation — 10 of 10 in the spec above, each missing `letterSpacing` and `lineHeight`.

`lineHeight` is the one that bites in practice: it's present in the source frontmatter and silently dropped, so a consumer round-tripping through DTCG loses it. (`letterSpacing` isn't in the DESIGN.md format at all, so it has no source value — worth deciding whether to emit a default, omit the `$type`, or target a profile that doesn't require it.)

### Suggested fixes, in order of preference

1. Emit `lineHeight` from the frontmatter, and emit `letterSpacing` with an explicit default (`"0"` / `{value: 0, unit: "px"}`) so the token validates.
2. If a required property genuinely has no source value, drop `$type: "typography"` for that token and emit the sub-values as a plain group — invalid-by-omission is worse than untyped.
3. Failing either, stop advertising the 2025.10 `$schema` in the output, so downstream validators don't reject a file that claims conformance.

Happy to send a PR for option 1 if that's the direction you'd want.

### Environment

`@google/design.md@0.4.0`, Node 25.2.1, macOS.

Arshgill01:
Fixes #172.

## What changed

`export --format dtcg` advertised the 2025.10 DTCG schema but omitted two of the five required typography properties (`lineHeight` and `letterSpacing`), so every typography token failed validation.

Two separate gaps:

1. **`lineHeight` was in the source but dropped.** YAML parses unquoted `lineHeight: 1.6` as a number. The typography model only accepted string dimensions, so the value never reached the DTCG emitter. Quoted strings (`"1.6"`) already worked. Finite numeric `lineHeight` is now kept as a unitless multiplier (the DTCG meaning). Bare numbers on `fontSize` / `letterSpacing` still error because those properties require an explicit unit.

2. **`letterSpacing` has no DESIGN.md source.** The DTCG 2025.10 typography value schema requires `letterSpacing`. When the role omits it, the exporter now emits `{ "value": 0, "unit": "px" }` so the token validates. Explicit source values are unchanged.

## How this differs from closed #174

#174 (unmerged) only kept YAML numeric `lineHeight` and **explicitly did not** invent a `letterSpacing` default. That left the schema-required field missing, so DTCG output still would not validate.

This PR includes that lineHeight model fix **and** emits the letterSpacing default so option 1 from #172 is complete.

## Verification

```bash
bun install
bun run lint
bun run test
bun run build
```

Targeted:

```bash
bun test src/linter/model/handler.test.ts src/linter/dtcg/handler.test.ts
```

## AI assistance

Cursor Grok assisted with investigation, TDD, and drafting this PR. I reviewed the change and ran the tests above.

### DIFF
--- a/packages/cli/src/linter/dtcg/handler.test.ts
+++ b/packages/cli/src/linter/dtcg/handler.test.ts
@@ -14,6 +14,7 @@
 
 import { describe, test, expect } from 'bun:test';
 import { DtcgEmitterHandler } from './handler.js';
+import { lint } from '../lint.js';
 import type { DesignSystemState, ResolvedColor, ResolvedDimension, ResolvedTypography } from '../model/spec.js';
 
 function emptyState(overrides?: Partial<DesignSystemState>): DesignSystemState {
@@ -170,7 +171,28 @@ describe('DtcgEmitterHandler', () => {
     expect(value['letterSpacing']).toEqual({ value: 0.5, unit: 'px' });
   });
 
-  test('typography with missing fields omits them from $value', () => {
+  test('typography without letterSpacing emits a 0px default', () => {
+    const body: ResolvedTypography = {
+      type: 'typography',
+      fontFamily: 'Manrope, Arial, sans-serif',
+      fontSize: makeDim(16, 'px'),
+      fontWeight: 400,
+      lineHeight: makeDim(1.6, ''),
+    };
+
+    const state = emptyState({
+      typography: new Map([['body-md', body]]),
+    });
+
+    const result = handler.execute(state);
+    expect(result.success).toBe(true);
+    if (!result.success) return;
+
+    const value = ((result.data['typography'] as Record<string, unknown>)['body-md'] as Record<string, unknown>)['$value'] as Record<string, unknown>;
+    expect(value['letterSpacing']).toEqual({ value: 0, unit: 'px' });
+  });
+
+  test('typography with missing fields omits them from $value except default letterSpacing', () => {
     const minimal: ResolvedTypography = {
       type: 'typography',
       fontFamily: 'Roboto',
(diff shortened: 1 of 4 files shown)


Arshgill01:
## Soft ping — workflow approval

Still **MERGEABLE** / `UNSTABLE` because **Test** (and the Google Actions Workflow Security Scan) are waiting on maintainer **Approve and run workflows** for this fork PR.

- CLA: pass
- GitHub Actions Scan: pass  
- Local + fork CI were green when opened

Run needing approval: https://github.com/google-labs-code/design.md/actions/runs/33625694230

Happy to rebase if `main` moves. Fixes #172.

